Issue #3, opened 2023-10-06
perf(syntax-highlight): cache highlight configuration


wongjiahau:
## Preface
When editing SQL files, I notice significant lag during typing. 

By using [Flamegraph](https://github.com/flamegraph-rs/flamegraph), I identified the bottleneck to be the construction of `HighlightConfiguration`:

![image](https://github.com/wongjiahau/treeman/assets/23183656/5f9ac953-6384-4709-8ff3-77d93c217b43)

## How is the profiling done?
1. Run `sudo cargo flamegraph`
2. Open an empty file that ends with `.sql`
3. Type in `-- who lives in a pineapple under the sea` character by character
4. Quit the editor

## References
- Raw flamegraph file (download and open to interact):
![flamegraph](https://github.com/wongjiahau/treeman/assets/23183656/d52650e3-ded1-46f6-b766-41d29f26d9a4)

wongjiahau:
Closed by #4

PR #4, opened 2023-10-06
perf(syntax-highlight): cache highlight configuration (#3)

wongjiahau:


wongjiahau:
The new performance profile shows that with these changes, `HighlightConfiguration::new()` is no longer the largest bottleneck.

flamegraph

PR #62, opened 2024-03-28
feat(syntax-highlight): fetch highlight queries from nvim-treesitter

wongjiahau:


wongjiahau:
## Note
The ordering of the queries is important, for example, in Typescript, if `(call_expression
  function: (identifier) @function.call)` is placed after `(identifier) @variable`, then `@function.call` will never be captured.

## Reference
- https://github.com/nvim-treesitter/nvim-treesitter/issues/3944#issuecomment-1455012371

wongjiahau:
Turns out the query precedence is reversed in https://github.com/tree-sitter/tree-sitter/pull/2947, (previously "first match wins", now "last match wins").

It means that for `nvim-treesitter` highlight queries to work properly, I only have to update `tree-sitter-highlight` crates to the latest version, since currently I'm using the version before the aforementioned PR.

PR #315, opened 2024-09-09
feat(lsp): zig

### LINKED ISSUE #314: feat(lsp): zig
Add lsp configuration for zig.

yves-biener:
closes #314

### DIFF
--- a/shared/src/languages.rs
+++ b/shared/src/languages.rs
@@ -374,7 +374,7 @@ const fn zig() -> Language {
             subpath: None,
             commit: "master",
         }),
-        formatter_command: Some(Command("zig", &["fmt"])),
+        formatter_command: Some(Command("zig", &["fmt", "--stdin"])),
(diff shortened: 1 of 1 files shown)


yves-biener:
there is currently one issue, that the formatting is not working as expected, as the format command requires the file path as an argument. How can I provide that parameter accordingly?

wongjiahau:
Variable substitution is not currently supported yet, but on the other hand, it seems like you can add the `--stdin` flag after `zig fmt` to make it work.

Reference: 
https://github.com/ziglang/zig/blob/d6d09f4ea7cc590035bf8af6617f93767423c691/lib/compiler/fmt.zig#L19

yves-biener:
You are right, thank you for the hint. I added the argument and tested it successfully 🎉

wongjiahau:
Thanks for your contribution!

PR #319, opened 2024-09-09
fix(editor): DisableMouseCapture on quit

mattdeeds:
Mouse capture with crossterm is enabled but never disabled on quit. On Mac this causes mouse events to be input to terminal until terminal is reset.


wongjiahau:
Thanks!

PR #324, opened 2024-09-11
Add LTO to the Release profile in the Cargo.toml

zamazan4ik:
Resolve https://github.com/ki-editor/ki-editor/discussions/321

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,3 +106,4 @@ indoc = "2.0.4"
 
 [profile.release]
 debug = true
+lto = true


wongjiahau:
Hey @zamazan4ik, thank you for your contribution! Would you be able to conduct benchmarking for both runtime performance and compile time to determine if LTO results in significant improvements?

wongjiahau:
Not sure if it's Placebo but it does feel much snappier with each keypress when LTO is enabled.

zamazan4ik:
> Would you be able to conduct benchmarking for both runtime performance and compile time to determine if LTO results in significant improvements?

Do you have an easy way to compare the project's performance between revisions? Maybe some kind of built-in benchmarks or smth like that? Because you know, measuring "with an eye" is not the most precise thing to do ;) Without that, I don't see performance lags from enabling LTO locally

wongjiahau:
Unfortunately, there's no easy way to do this, but what I usually do is run `sudo cargo flamegraph` in the project root. Then, I open a file and perform an insertion or modification and exit the editor, then you'll get a `flamegraph.svg`.

After that, repeat the same action on a different branch, then compare the two flamegraphs.

wongjiahau:
@Riolku what do you think of this change? Theoretically it will improve runtime performances via aggresive optimizations, but it can also increase build time significantly.

Should we close or merge this PR?

Riolku:
10% was the speedup i got in a C++ project some years ago

PR #328, opened 2024-09-13
fix: url conversion

Vulpesx:
fix bad url conversion to string that resulted in bad file paths


wongjiahau:
Thanks! Can you add some unit test cases?

wongjiahau:
Staled

PR #331, opened 2024-09-13
feat(style): show last dispatch global bar

### LINKED ISSUE #256: style(global bar): show last dispatch
This is to aid learning.

yves-biener:
closes #256

### DIFF
--- a/src/components/editor.rs
+++ b/src/components/editor.rs
@@ -2908,10 +2908,10 @@ impl std::fmt::Display for DispatchEditor {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Surround(open, close) => write!(f, "Surround {open}{close}",),
-            ShowJumps { .. } => write!(f, "{}", "Show Jumps"),
-            ScrollPageDown => write!(f, "{}", "Scroll Page Down"),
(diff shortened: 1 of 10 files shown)


src/components/editor.rs
```diff
@@ -2892,6 +2904,109 @@ pub(crate) enum DispatchEditor {
     Dedent,
 }
 
+impl std::fmt::Display for DispatchEditor {
```
yves-biener, line 2907: I'm not sure if we should use a proc_macro for this or keep it that way. When using debug printing it will display other information alongside the enum variant name (i.e. parameters like `use_system_clipboard`) which might not be desired, as it will fill up the global bar.

wongjiahau: I think this code is fine for now, but eventually, I want to have an extra line for the status bar, so we can show the full dispatch details in the second line without having to be afraid that it will overflow.

yves-biener: There might be need to overhaul the bar as a hole later anyway, where we can change this too, as advanced user might want to disable this feature anyway. Do you want to keep this in mind or create another issue for later?

wongjiahau:
Hey @yves-biener, just checking in to see if you would like to complete the remaining parts of this PR?

wongjiahau:
Thanks @yves-biener great work!

Issue #332, opened 2024-09-14
Keybinding help with `@n` panics


ilyagr:
After looking through the source code, I was excited to discover that the `@` keybinding is supposed to show keybinding help. However, pressing `@n` for normal mode bindings resulted in a panic:

```
thread 'main' panicked at src/components/keymap_legend.rs:235:26:
                                                                   called `Result::unwrap()` on an `Err` value: Syntax(
                                                                                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                              regex parse error:
                                                                                                    ( → Cycle primary selection (?<hint>\()backward\)
                                 ^
                                  error: unclosed group
                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
              )
               note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I suspect the intent was to disable this feature for now, until it works better (since it doesn't seem to be documented).

As an aside, I think having this feature work and advertised prominently (e.g. as `:help` and/or in the Space menu) would be very helpful.

wongjiahau:
Yup, it’s undocumented because @ is a weird keybinding, what do you think of `space ?`?

ilyagr:
`space ?` sounds great to me. I vaguely remember that Spacemacs or Doom Emacs might have used that.

wongjiahau:
Do you want to give it a try?

ilyagr:
Maybe at some point, but no promises. I'm more worried about the panic than the binding, by the way.

ilyagr:
OK, I did find a quick fix to the panic in #334 :). I'm not sure it solves the issue completely.

I'm still not sure this "works": how do I scroll through the resulting window after I press `@n`?

wongjiahau:
> I'm still not sure this "works": how do I scroll through the resulting window after I press `@n`?

The answer lies in "Every component is an editor".

After you press `@n`, you can press `esc` to enter normal mode, and then use `ctrl+d`/`ctrl+u` to scroll down and up.

If you want to execute any keybindings, enter insert mode (either `i`/`a`), then press the key.

ilyagr:
Ah, thanks!

[...]

PR #334, opened 2024-09-17
fix panic on keybinding help, `@n`

ilyagr:
This is a quick fix to the panic in #332.

Still TODO; change the key binding, document it, perhaps investigate this issue in more detail.


wongjiahau:
Thanks!

PR #361, opened 2024-10-04
Nix flake

mjakobsche:
Hey,
I created this flake, which allows users to install ki-editor using nix. It also defines an environment with basic packages for development . If there is something I should add or fix, let me know. 
I think it will encourage more people to contribute ;> and make the installation process even more simple.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/result
 /target
 tree_sitter_quickfix/build/
 tree_sitter_quickfix/node_modules/
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -1,7 +1,5 @@
 # Installation
 
-Currently, only the "build from source" method is available.
-
 ## Build from source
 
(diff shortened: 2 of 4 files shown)


wongjiahau:
I’m going to trust you on this one as I’m not well-versed in Nix.

Thanks for the contribution @mjakobsche!

evelant:
Unfortunately this does not compile for me on mac (aarch64-darwin). It fails with a linker error building libz-sys `note: ld: library not found for -liconv`. A missing dep?

mjakobsche:
You are correct. I will fix it.

mjakobsche:
@evelant could you let me know if this fixes the issue:
`buildInputs = with pkgs; [
            openssl
          ] ++ lib.optionals pkgs.stdenv.isDarwin [
            libiconv
            curl
            darwin.apple_sdk.frameworks.AppKit
          ]; `

PR #455, opened 2024-12-07
"ki edit ." now opens file explorer instead of crashing.

AlphabetsAlphabets:
It used to crash at `main.rs` where `cli::cli.unwrap()` was called because you'd just get an `Err` as the `edit` match arm wasn't able to handle a directory instead of a file. The file explorer does open but it is slow. I'm wondering if it's an issue of how I opened the file explorer or if it's an issue of the file explorer.

There is a bug where `ki edit ..` opens the current directory instead of the parent directory (will work on it) but I thought to just open the PR so it is kept track of it.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2,16 +2,11 @@ use crate::{
     buffer::Buffer,
     clipboard::CopiedTexts,
     components::{
-        component::{Component, ComponentId, GetGridResult},
-        dropdown::{DropdownItem, DropdownRender},
-        editor::{DispatchEditor, Editor, IfCurrentNotFound, Movement},
-        keymap_legend::{
+        component::{Component, ComponentId, GetGridResult}, dropdown::{DropdownItem, DropdownRender}, editor::{DispatchEditor, Editor, IfCurrentNotFound, Movement}, file_explorer::FileExplorer, keymap_legend::{
             Keymap, KeymapLegendBody, KeymapLegendConfig, KeymapLegendSection, Keymaps,
-        },
-        prompt::{Prompt, PromptConfig, PromptHistoryKey},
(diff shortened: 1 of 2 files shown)


wongjiahau:
Thanks @AlphabetsAlphabets! Can you fix the Github actions error by using `rustfmt 1.7.0-stable (05147895 2024-07-21)`?

AlphabetsAlphabets:
Yes, I can do that. But I'm still worried about the performance of the File Explorer. It's a bit slow, and I'm not sure if my way of invoking it is the right way. Unless the explorer is just slower which I don't think it is.

Sent from Proton Mail Android

-------- Original Message --------

wongjiahau:
I don't think your code is causing the performance issue, but if I were pedantic, the file explorer should be opened in the directory passed from the CLI arguments.

AlphabetsAlphabets:
@wongjiahau I've done the formatting with rustfmt, feel free to push. But the bug of opening the current dir when `ki edit ..` is still there. 

I'm trying my best to fix it but I'm a little confused and would like some help. I'll save the yapping lol but basically, I'd like to know your debugging setup. If you think it's better to open a discussion, I'll do just that. I'm also in the matrix chat if you prefer that I DM you there, either way is fine.

wongjiahau:
Yea sure! Let’s discuss on Matrix, can you DM me?

AlphabetsAlphabets:
Thanks :) I'll send my question over. I won't be able to respond immediately though just FYi

Sent from Proton Mail Android

-------- Original Message --------

AlphabetsAlphabets:
@wongjiahau hi, I managed to fix the bug with "ki edit .." opens the cwd instead of parent.

Issue #463, opened 2024-12-18
bug(cli): `ki edit` not working on files now


wongjiahau:
Potentially due to 29f80df3ac43b971966484ab3157f7d64d1fe6d8

PR #466, opened 2024-12-20
fix(cli): ki edit now works on files as well as . and .. (#463)

jcowgar:
In the call to `RunConfig`, `working_directory` must be a directory. The bug was that it was a file.

With this patch, if the path supplied on the CLI is a file, `working_directory` will become the parent of whatever was specified on the command line. If it is not a file, it is taken as is. If not specified, it becomes `Default::default()`.


wongjiahau:
Thanks for your contributions!

PR #469, opened 2024-12-21
fix for c, added support for cpp

Seres67:


Seres67:
maybe add support for more filetypes for cpp? [here](https://github.com/helix-editor/helix/blob/master/languages.toml#L548) is the list helix uses

Issue #473, opened 2024-12-24
feat(lsp): support DocumentSymbol


wongjiahau:
LSP servers can return symbol information as SymbolInformation or DocumentSymbol. SymbolInformation is a flat list while DocumentSymbol can contain children and will self reference.

So far, in our Language Definitions, the Gleam and Swift LSP servers return DocumentSymbol. The Pick Symbol feature fails to function for these LSP servers.

Beside the nested functionality of DocumentSymbol, the other big change is that DocumentSymbol does not contain any file reference like SymbolInformation does. Without a reference to the file, the symbol picker in Ki does not know which file the symbol lives in and will fail to work.

Reference: 

* [DocumentSymbol specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol)

PR #475, opened 2024-12-24
feat(lsp): support DocumentSymbol and children

jcowgar:
Solves #473.

* Add support for parsing `Nested` in addition to `Flat` responses from LSP servers for `Document Symbol`.
* `DocumentSymbol` does not have file path information, therefore:
  * Introduce new `go_to_current_component_range` function
  * Introduce new `GoToCurrentComponentRange` `Dispatch` type. This type only has a range and will use the current component.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2527,18 +2527,17 @@ impl DispatchPrompt {
                 // so that we don't have to do this,
                 // i.e. we can just return the symbol directly,
                 // instead of having to find it again.
-                let dispatches: Dispatches = match symbols
+                if let Some(symbol) = symbols
                     .symbols
                     .iter()
                     .find(|symbol| text == symbol.display())
                 {
-                    Some(symbol) => {
-                        Dispatches::one(Dispatch::GotoLocation(symbol.location.to_owned()))
-                    }
-                    None => Dispatches::new(vec![]),
-                };
-
-                Ok(dispatches)
+                    Ok(Dispatches::new(vec![Dispatch::GotoLocation(
+                        symbol.location.clone(),
+                    )]))
+                } else {
+                    Ok(Dispatches::new(vec![]))
+                }
             }
             DispatchPrompt::RunCommand => Ok(Dispatches::new(
                 [Dispatch::RunCommand(text.to_string())]
--- a/src/app.rs
+++ b/src/app.rs
@@ -1171,12 +1171,7 @@ impl<T: Frontend> App<T> {
     }
 
     fn go_to_location(&mut self, Location { path, range }: &Location) -> Result<(), anyhow::Error> {
-        let component = if *path == "/".try_into()? {
-            self.current_component()
-        } else {
-            self.open_file(path, OpenFileOption::Focus)?
-        };
-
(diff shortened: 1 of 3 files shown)


jcowgar:
Currently, children DocumentSymbols are added into the mix at the top level, so no displayed nesting yet.

src/app.rs
wongjiahau, line 1: The changes here seems unnecessary at this point

src/lsp/symbols.rs
```diff
@@ -4,45 +4,84 @@ use crate::{
...
 }
 
-impl TryFrom<DocumentSymbolResponse> for Symbols {
-    type Error = anyhow::Error;
+fn collect_document_symbols(
+    document_symbol: &lsp_types::DocumentSymbol,
+    symbols: &mut Vec<Symbol>,
+    parent_name: Option<String>,
+    path: &CanonicalizedPath,
+) -> Result<(), anyhow::Error> {
```
wongjiahau, line 19: I would prefer this function to return a vector instead of taking a mutable vector parameter, as that would be more composable and would result in fewer changes in this PR.

src/lsp/symbols.rs
```diff
@@ -65,11 +104,15 @@ impl Symbol {
...
-                symbol.container_name.unwrap_or("[TOP LEVEL]".to_string()),
+                symbol
+                    .container_name
+                    .clone()
+                    .unwrap_or("[TOP LEVEL]".to_string()),
             ))
-            .set_dispatches(dispatches)
+            .set_dispatches(Dispatches::one(Dispatch::GotoLocation(
+                symbol.location.to_owned(),
+            )))
```
wongjiahau, line 116: The changes here should no longer be necessary.

src/lsp/symbols.rs
```diff
@@ -4,45 +4,84 @@ use crate::{
...
 
-impl TryFrom<lsp_types::SymbolInformation> for Symbol {
-    type Error = anyhow::Error;
-
-    fn try_from(value: lsp_types::SymbolInformation) -> Result<Self, Self::Error> {
+impl Symbol {
+    fn try_from_symbol_information(value: lsp_types::SymbolInformation) -> anyhow::Result<Self> {
         let name = value.name;
-        let location = value.location.try_into()?;
+        let location = Location::try_from(value.location)?;
```
wongjiahau, line 59: My bad, this change here is unnecessarily committed.

src/lsp/symbols.rs
```diff
@@ -4,45 +4,84 @@ use crate::{
...
+        for child in children {
+            collect_document_symbols(&child, symbols, Some(document_symbol.name.clone()), path)?;
+        }
+    };
+
+    Ok(())
+}
+
+impl Symbols {
+    pub(crate) fn try_from_document_symbol_response(
```
wongjiahau, line 34: Making `collect_document_symbols` pure would reduce the necessary changes in this function.

wongjiahau (changes requested):

wongjiahau (approved):

PR #478, opened 2024-12-26
feat(keymap/space): add commands to space menu

### LINKED ISSUE #477: Make commands available on Space Menu
`:` access the [Command Mode](https://ki-editor.github.io/ki-editor/docs/normal-mode/commands). Currently there are only three commands, 

* Quit All
* Write All
* Write Quit All

These would be more accessible via the [Space Menu](https://ki-editor.github.io/ki-editor/docs/normal-mode/space-menu).

jcowgar:
Add Write All, Write Quit All and Quit All commands to the space menu.

Command mode was not touched, it still remains unchanged. This could be the first step in getting away from it. Next step could be to simply unbind the ":" key and third step removing the code, if desired.

resolves #477

### DIFF
--- a/src/components/editor_keymap_legend.rs
+++ b/src/components/editor_keymap_legend.rs
@@ -931,21 +931,13 @@ impl Editor {
                     .chain(Some(KeymapLegendSection {
                         title: "File/Quitting".to_string(),
                         keymaps: Keymaps::new(&[
-                            Keymap::new(
-                                "w",
-                                "Write All".to_string(),
-                                Dispatch::RunCommand("write-all".to_string()),
-                            ),
+                            Keymap::new("w", "Write All".to_string(), Dispatch::SaveAll),
                             Keymap::new(
                                 "q",
                                 "Write All and Quit".to_string(),
-                                Dispatch::RunCommand("write-quit-all".to_string()),
+                                Dispatch::SaveQuitAll,
                             ),
-                            Keymap::new(
-                                "Q",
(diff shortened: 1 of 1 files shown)


wongjiahau (changes requested):
I think we can remove `Dispatch::RunCommand` and create 3 new variants under `Dispatch` for these 3 operations.

jcowgar:
> I think we can remove `Dispatch::RunCommand` and create 3 new variants under `Dispatch` for these 3 operations.

That would mean getting rid of command mode all together, correct? Do we want to take that step right away?

wongjiahau:
Yes, let’s remove Command mode as well

wongjiahau:
Or do you have any concerns with removing it?

wongjiahau:
By the way for future PRs, what do you think about creating the branch in this repository so that I may work on it directly as well?

jcowgar:
OK, I'll remove it. And yes, I'll create branches here in this repo. Wasn't sure if you wanted me doing that or not.

jcowgar:
@wongjiahau turns out the dispatch commands already existed, so I simply updated the menu to call them instead of RunCommand.

What do you think about this PR standing as is and we create a new ticket to remove Command Mode? As this PR doesn't have to remove command mode and if we ever want to bring it back for some reason, we would then have a single commit that we could revert?

wongjiahau:
Ok! Let’s keep the scope of this PR as true to its intent as possible

PR #483, opened 2024-12-26
feat(keymap): add previous/next buffer navigation

jcowgar:
Will cycle through currently open file buffers. If you get to the end of the buffer list, it will start over at the beginning. Likewise, if you are at the start of the buffer list and go previous, it will cycle back to the last buffer.

When you close a buffer with `ctrl+c`, it will no longer be part of the buffer navigation.

Currently, this functionality is bound to `-` and `=`. Better keys may be `tab` and `shift+tab` but that has a conflict currently with Go back and Go forward.

This is different than previous/next opened files as that will cycle through files that you opened at some point in time. Closing a file does not remove it from the history stack.

### DIFF
--- a/docs/docs/normal-mode/other-movements.md
+++ b/docs/docs/normal-mode/other-movements.md
@@ -60,3 +60,10 @@ Keybindings:
 
 - `{`: Go to previously opened file
 - `}`: Go to the next opened file
+
+## Go to the previous/next buffer
+
+Keybindings:
+
+- `-`: Go to previous buffer
+- `=`: Go to next buffer
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,7 +4,7 @@ use crate::{
     components::{
         component::{Component, ComponentId, GetGridResult},
         dropdown::{DropdownItem, DropdownRender},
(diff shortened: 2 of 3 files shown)


wongjiahau:
This sounds like a better version of Goto next/prev file, should we instead replace it with Goto next/prev buffer?

Or do you have use case for Goto next/prev file that is not covered by Goto next/prev buffer?

jcowgar:
I think that I will certainly use this more than the existing file prev/next functionality. I am not certain where it comes into play. The thing that I disliked most about it is when I am doing a global search, browsing through files learning code, searching for that exact place I need to edit, etc... all of those files were put onto the history stack. Now I can't get back easily to the 1 or 2 files that I am really interested in.

Hence, closing files and simply having prev/next keys to cycle through open buffers.

I could see binding this to the previous/next file keys and moving the previous/next file functionality to -/=. However, I wonder if we should leave the file prev/next functionality for a while until we see which one is really used the most?

Maybe even not worry about changing the bindings, keeping the -/= for buffer and then addressing in the upcoming positional key mapping, when that is done.

wongjiahau (approved):

Issue #484, opened 2024-12-26
Convert cli sub-commands to flags


jcowgar:
New usage. Edit will be removed and become the default.

```
Usage: ki [args] file...

Options:

--root                  Set the root path for the Ki session
--grammar-fetch         Fetch the latest tree-sitter grammar modules
--grammar-build         Builds the tree-sitter grammar modules
--highlight-query-clean Clean cached highlight queries
--highlight-query-path  Print the path for the highlight queries
--print-log-path        Print the path to the Ki log file
-h, --help              Print help
-V, --version           Print version
```

Current usage:

```
Usage: ki [COMMAND]

Commands:
  grammar
  highlight-query
  edit             Edit the file of the given path, creates a new file at the path if not exist
  log              Prints the log file path
  in               Run Ki in the given path, treating the path as the working directory
  help             Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

wongjiahau:
Now I that looked at it I think we should just nest the existing subcommands under a flag (perhaps --run), because flattening the commands into flags make it seemed like they can be used together but in fact they are mutually exclusive.

jcowgar:
That is true... It does present a little UX inconsistency. Just looked at helix:

```
~/Projects/ki-editor master ❱ hx -g fetch readme.md
Fetching 214 grammars
214 up to date git grammars
~/Projects/ki-editor master ❱
```

What would a user expect there? Maybe to fetch the grammar files and then load readme.md? That's not what it did though, which is exactly what you are saying.

jcowgar:
I think I have this solved another way. Please take a peek at the PR #488. If accepted, I'll update the ticket detail.

PR #488, opened 2024-12-27
feat(cli): make edit the default command

jcowgar:
See #484.

This PR hides the "edit" command and makes it default.

Examples are given at the root of the ki-language git repo.

* `ki` -- opens [ROOT] window as it always has
* `ki readme.md` -- opens `readme.md`
* `ki src` -- opens the File Explorer for the `src` directory
* `ki log` -- opens a new file named `log`
* `ki @ log` -- prints log path and exits
* `ki @ grammar` -- shows grammar help and exits
* `ki @ grammar build` -- builds grammar and exits

New help output from `ki --help`:

```
Usage: ki [PATH] [COMMAND]

Commands:
  @     Run commands
  help  Print this message or the help of the given subcommand(s)

Arguments:
  [PATH]

Options:
  -h, --help     Print help
  -V, --version  Print version
```

Output of `ki @ --help`

```
Run commands

Usage: ki @ <COMMAND>

Commands:
  grammar          Build and fetch tree-sitter grammar files
  highlight-query  Manage cached tree-sitter highlight files
  log              Prints the log file path
  in               Run Ki in the given path, treating the path as the working directory
  help             Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2614,6 +2614,7 @@ name = "shared"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "etcetera",
  "fancy-regex",
  "grammar",
  "isahc",
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,6 +6,9 @@ use shared::canonicalized_path::CanonicalizedPath;
 struct Cli {
     #[command(subcommand)]
     command: Option<CommandPlaceholder>,
+
+    #[command(flatten)]
+    edit: EditArgs,
 }
 
 #[derive(Subcommand)]
@@ -124,6 +127,6 @@ pub(crate) fn cli() -> anyhow::Result<()> {
                 ..Default::default()
             }),
         },
-        None => crate::run(Default::default()),
+        None => run_edit_command(cli.edit),
     }
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -4,13 +4,23 @@ use shared::canonicalized_path::CanonicalizedPath;
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
 struct Cli {
-    /// Run a command, such as grammar, highlight-query, ...
-    #[arg(long, action=clap::ArgAction::SetTrue)]
-    run: bool,
     #[command(subcommand)]
-    command: Option<Commands>,
-    #[command(flatten)]
+    command: Option<CommandPlaceholder>,
     edit: EditArgs,
+}
+
+#[derive(Subcommand)]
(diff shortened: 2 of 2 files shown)


lydell:
If I have a folder called “log” or “grammar”, is it possible to open the File Explorer for either of those folders? Is it `ki ./log` and `ki ./grammar` maybe?

jcowgar:
> If I have a folder called “log” or “grammar”, is it possible to open the File Explorer for either of those folders? Is it `ki ./log` and `ki ./grammar` maybe?

That is exactly how it will work with this PR.

You can, however, do it right now, prior to this PR, by `ki edit ./grammar` or `ki edit ./log`.

jcowgar:
I just thought of a conflict here, maybe it is what @wongjiahau was thinking of to begin with but say you have a directory named `log`, and you type `ki log`. It would execute the `log` command printing the path to the log file. It wouldn't open the `log` directory like `ki src` would.

This will happen with grammar, highlight-query, log, in and help.

I think this is an acceptable state to make using ki much easier. I think that the number of times you will try to edit a directory named log, grammar, highlight-query or help will be minimal.

You can work around this by prefixing your file path with a `./`, i.e. `ki ./log` will open the log directory in ki just fine, not launch the log command.

wongjiahau:
I actually name a log file as `log` quite a lot, but regardless I think the subcommands need to be nested for the consistency and predictability, otherwise it will become like `npm` where you are never sure what xxx is in `npm xxx`, as it could be either a built-in command or user-defined comnand.

jcowgar:
> I actually name log files as `log` quite a lot, but regardless I think the subcommands need to be nested for the consistency and predictability, otherwise it will become like `npm` where you are never sure what xxx is in `npm xxx`, as it could be either a built-in command or user-defined comnand.

I'm not sure what this means. Can you give me an example what you mean about subcommands being nested?

wongjiahau:
Yea sure, for example:

```
ki --run log
ki --run grammar fetch
```

jcowgar:
I understand now. I'll make the necessary changes.

wongjiahau:
I've decided to take compromises by nesting the existing subcommands under the `@` subcommand, this offers the best of both worlds:
1. It allows `ki <filename>` to work
2. Its implementation is not unconventional

The only drawback is that if a user ever wants to open a file named `@`, he has to do `ki ./@`, sort of like escaping backslash in a string literal.

wongjiahau (approved):

PR #491, opened 2024-12-27
feat(explorer): add copy file to space menu for explorer

jcowgar:
Copy a single file from one location to another location.

* Added to explorer space context menu
* Added documentation
* Added notification to the LSP server that a new file was created

### DIFF
--- a/docs/docs/normal-mode/space-menu.md
+++ b/docs/docs/normal-mode/space-menu.md
@@ -26,6 +26,7 @@ Contextual actions are actions that are only applicable within a specific contex
 | Keybinding | Action                                                           |
 | ---------- | ---------------------------------------------------------------- |
 | `a`        | Add a new file/folder under the current path [^1]                |
+| `c`        | Copy current file to a new file                                  |
 | `d`        | Delete current file/folder                                       |
 | `m`        | Move (or rename) the current file/folder [^2]                    |
 | `r`        | Refresh the [file explorer](../components/file-explorer.md) [^3] |
--- a/docs/docs/normal-mode/space-menu.md
+++ b/docs/docs/normal-mode/space-menu.md
@@ -26,7 +26,7 @@ Contextual actions are actions that are only applicable within a specific contex
 | Keybinding | Action                                                           |
 | ---------- | ---------------------------------------------------------------- |
 | `a`        | Add a new file/folder under the current path [^1]                |
-| `c`        | Copy current file to a new file                                  |
+| `c`        | Copy current file to a new path                                  |
 | `d`        | Delete current file/folder                                       |
 | `m`        | Move (or rename) the current file/folder [^2]                    |
 | `r`        | Refresh the [file explorer](../components/file-explorer.md) [^3] |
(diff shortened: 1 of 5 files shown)


docs/docs/normal-mode/space-menu.md
```diff
@@ -26,6 +26,7 @@ Contextual actions are actions that are only applicable within a specific contex
 | Keybinding | Action                                                           |
 | ---------- | ---------------------------------------------------------------- |
 | `a`        | Add a new file/folder under the current path [^1]                |
+| `c`        | Copy current file to a new file                                  |
```
wongjiahau, line 29: We should use `y` since it stands for Copy at the moment.

Also, I think `Copy current file to a new path` is more clear.

src/app.rs
```diff
@@ -864,6 +866,21 @@ impl<T: Frontend> App<T> {
         )
     }
 
+    fn open_copy_file_prompt(&mut self, path: CanonicalizedPath) -> anyhow::Result<()> {
+        self.open_prompt(
+            PromptConfig {
+                title: "Copy path".to_string(),
```
wongjiahau, line 872: I think it would be clearer if the title is formatted like this:

`Copy <current file path> to:`

src/app.rs
```diff
@@ -1348,6 +1365,21 @@ impl<T: Frontend> App<T> {
...
+    fn copy_file(&mut self, from: CanonicalizedPath, to: PathBuf) -> anyhow::Result<()> {
+        use std::fs;
+        self.add_path_parent(&to)?;
+        fs::copy(from.clone(), to.clone())?;
+        self.layout.refresh_file_explorer(&self.working_directory)?;
+        let to = to.try_into()?;
+        self.reveal_path_in_explorer(&to)?;
+        self.lsp_manager.send_message(
+            from.clone(),
+            FromEditor::WorkspaceDidCreateFiles { file_path: to },
```
wongjiahau, line 1378: Nice! Can you add this to `add_path` as well?

wongjiahau (changes requested):

docs/docs/normal-mode/space-menu.md
jcowgar: y to me means yank, put it on the clipboard for example. In most File Managers you can "copy" a file and paste it somewhere else.

I think of this copy as an immediate action, hence different.

What do you think?

src/app.rs
jcowgar: Do you mean put in the actual text of the file being copied?

jcowgar: Looking at the current menu, options are:

* a -> Add File (or postfix with / for folder)
* d -> Delete path
* m -> Move path

They do not display anything about the current path.

docs/docs/normal-mode/space-menu.md
wongjiahau: Make sense

wongjiahau (approved):

wongjiahau:
Is this ready to be merged?

jcowgar:
Yes, I enabled auto merge and merged in master. It's building.

PR #501, opened 2025-01-01
feat(explorer): Open multiple files with one action

jcowgar:
`ctrl+o` will open all files in the Completion list that you have narrowed down with the file picker.

### DIFF
--- a/src/components/prompt.rs
+++ b/src/components/prompt.rs
@@ -5,7 +5,7 @@ use my_proc_macros::key;
 use crate::{
     app::{Dispatch, DispatchPrompt, Dispatches, GlobalSearchFilterGlob, Scope},
     buffer::Buffer,
-    components::{editor::DispatchEditor, keymap_legend::Keymaps},
+    components::editor::DispatchEditor,
     context::Context,
     lsp::completion::Completion,
     selection::SelectionMode,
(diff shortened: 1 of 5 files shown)


src/components/prompt.rs
```diff
@@ -127,22 +127,30 @@ impl Component for Prompt {
...
+
     fn handle_key_event(
         &mut self,
         context: &Context,
         event: event::KeyEvent,
     ) -> anyhow::Result<Dispatches> {
         match event {
+            key!("ctrl+o") if self.prompt_history_key == PromptHistoryKey::OpenFile => {
+                let mut final_dispatches = Dispatches::new(vec![Dispatch::CloseCurrentWindow]);
+                final_dispatches.extend(self.editor.completion_all_filtered_dispatches());
```
wongjiahau: You should use `Dispatches::chain` instead ti minimize usage of `mut`.

src/components/suggestive_editor.rs
```diff
@@ -203,6 +202,21 @@ impl SuggestiveEditor {
...
+        self.completion_dropdown
+            .filtered_item_groups()
+            .iter()
+            .for_each(|i| {
+                i.items().iter().for_each(|ci| {
+                    final_dispatches.extend(ci.item().dispatches);
+                })
+            });
+
+        final_dispatches
```
wongjiahau: Please use `flat_map` and `map` instead.

src/components/suggestive_editor.rs
```diff
@@ -203,6 +202,21 @@ impl SuggestiveEditor {
         self.completion_dropdown.current_item()
     }
 
+    pub(crate) fn completion_all_filtered_dispatches(&self) -> Dispatches {
+        let mut final_dispatches = Dispatches::new(vec![]);
+
+        self.completion_dropdown
+            .filtered_item_groups()
+            .iter()
+            .for_each(|i| {
```
wongjiahau: Please avoid abbreviations for naming lambda parameter, for example:

Do:
```
items.map(|item| item.foo())
```

Don’t:
```
items.map(|i| i.foo())
```

wongjiahau (changes requested):
The implementation is as I expected, only minor coding convention changes are required. 

 
Just curious, is this working as you expect?

wongjiahau (approved):

PR #502, opened 2025-01-02
feat(editor): dirty detection for saving

jcowgar:
Currently when you press `enter` or exit the editor with `space q`, the file is saved always. This has a few undesired side effects:

1. If the file has LSP attached, it is notified the file has changed and work is done for no reason.
2. If the file has a formatter, it is called in an attempt to format the code. Again, work done for no reason.
3. If the file changes external to the editor, your external changes are blindly overwritten with the content in your buffer which you made no changes to.

If you have a long editing session, with many open files and a slow formatter, even quitting the editor can take a considerable time. If no changes are made, exiting is instant.

With this change, save is only attempted if the editor has been changed since it was read or last saved.

### DIFF
--- a/docs/docs/normal-mode/space-menu.md
+++ b/docs/docs/normal-mode/space-menu.md
@@ -3,6 +3,7 @@
 The space menu is a handy shortcut for (not restricted to):
 
 - Contextual actions
+- File and quit actions
 - Searching files/symbols
 - Multi-cursor management
 - Opening other components
@@ -12,6 +13,15 @@ The space menu can be brought up by pressing `space`.
 ## Contextual actions
 
 Contextual actions are actions that are only applicable within a specific context.
+
+## File and quit actions
+
+| Keybinding | Action                                           |
+| ---------- | ------------------------------------------------ |
+| `space`    | Write current file, even if no changes were made |
+| `w`        | Write all files                                  |
+| `q`        | Write all files and quit                         |
+| `Q`        | Quit _without_ writing files                     |
 
 ### LSP Actions (only applicable in the main editor):
 
--- a/docs/docs/normal-mode/space-menu.md
+++ b/docs/docs/normal-mode/space-menu.md
@@ -21,7 +21,7 @@ Contextual actions are actions that are only applicable within a specific contex
 | `space`    | Write current file, even if no changes were made |
 | `w`        | Write all files                                  |
 | `q`        | Write all files and quit                         |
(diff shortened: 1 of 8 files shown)


src/components/test_editor.rs
```diff
@@ -1979,6 +1979,34 @@ fn entering_insert_mode_from_visual_mode() -> anyhow::Result<()> {
...
 }
 
+#[test]
+fn modifying_editor_causes_dirty_state() -> anyhow::Result<()> {
+    execute_test(|s| {
+        Box::new([
+            App(OpenFile(s.main_rs())),
+            Expect(Not(Box::new(EditorIsDirty()))),
+            Editor(EnterInsertMode(Direction::Start)),
+            App(HandleKeyEvents(keys!("a a esc").to_vec())),
```
wongjiahau, line 1989: Better use `DispatchEditor`s rather than `keys!` so that these test cases are not affected by the keymap changes.

src/context.rs
wongjiahau, line 1: I think the changes to this file should be in a separate PR.

src/components/test_editor.rs
wongjiahau: Oops, I didn't notice that's in Insert mode.

src/components/editor.rs
```diff
@@ -498,6 +498,7 @@ pub(crate) enum Movement {
 
 impl Editor {
     /// Returns if the editor is currently dirty or not.
+    #[test]
```
wongjiahau, line 501: This should be `#[cfg(test)]`

docs/docs/normal-mode/space-menu.md
```diff
@@ -13,6 +14,15 @@ The space menu can be brought up by pressing `space`.
...
 Contextual actions are actions that are only applicable within a specific context.
 
+## File and quit actions
+
+| Keybinding | Action                                           |
+| ---------- | ------------------------------------------------ |
+| `space`    | Write current file, even if no changes were made |
+| `w`        | Write all files                                  |
+| `q`        | Write all files and quit                         |
+| `Q`        | Quit _without_ writing files                     |
```
wongjiahau, line 24: I think this can be phrased better as:
```md
Quit _without_ writing unsaved files
```

wongjiahau (changes requested):
Looks good to me except a minor change in the docs, feel free to merge this PR after you've updated the docs!

wongjiahau (approved):

PR #503, opened 2025-01-03
Refactor code for selecting/assigning themes

jcowgar:
Simplify and ease maintainability of the Theme selection by introducing a `ThemeDescriptor`. This allows us to define a theme name and a method of instantiating that theme when needed. Future theme types can easily be added.

This also has the side effect of lazily fetching and instantiating themes. Therefore, saving time and memory. This side effect has taken our tests (on my M3 MacBook Air) from 62s to 9s.

As part of this, while changing various parts of code two additional "features" were implemented:

1. Added One theme from the Zed repo.
2. Ordered Alphabetized Theme's in the Theme Selector


src/themes/from_zed_theme.rs
```diff
@@ -1,210 +1,285 @@
...
+        ThemeDescriptor::ZedThemeURLMap("Atelier Plateau Light", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Savanna Dark", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Savanna Light", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Seaside Dark", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Seaside Light", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Sulphurpool Dark", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Atelier Sulphurpool Light", ZED_THEME_ATELIER_URL),
+        ThemeDescriptor::ZedThemeURLMap("Ayu Dark", ZED_THEME_AYU_URL),
+        ThemeDescriptor::ZedThemeURLMap("Ayu Light", ZED_THEME_AYU_URL),
+        ThemeDescriptor::ZedThemeURLMap("Ayu Mirage", ZED_THEME_AYU_URL),
```
wongjiahau, line 61: We can reduce the repetition of `ThemeDescriptor::ZedThemeURLMap` if we pack the arguments into tuples:
```rs
[
   // ...
   ("Ayu Dark", ZED_THEME_AYU_URL),
   ("Ayu Mirage", ZED_THEME_AYU_URL),
   // ...
]
    .into_iter()
    .map(ThemeDescriptor::ZedThemeURLMap) 
    // If point free style doesn't work due to borrowing, do this:
    // .map(|(name, url)| ThemeDescriptor::ZedThemeURLMap(name, url))
    .collect_vec()
```

src/themes/theme_descriptor.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut theme_descriptors: Vec<ThemeDescriptor> = [
+        ThemeDescriptor::ThemeFn("VS Code (Light)".to_string(), vscode_light),
+        ThemeDescriptor::ThemeFn("VS Code (Dark)".to_string(), vscode_dark),
+    ]
+    .into_iter()
+    .chain(from_zed_theme::theme_descriptors())
+    .collect();
+
+    theme_descriptors.sort_by(|a, b| a.name().cmp(b.name()));
+    theme_descriptors
```
wongjiahau, line 41: You can import  `itertools::Itertools`, and then use [`sorted_by_key`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.sorted_by_key) to avoid `mut` here:
```rs
// ...
    .into_iter()
    // ...
    .sorted_by_key(|theme_descriptor| theme_descriptor.name())
    // ... 
```

src/app.rs
```diff
@@ -2071,11 +2071,13 @@ impl<T: Frontend> App<T> {
...
-                items: crate::themes::themes()?
+                items: crate::themes::theme_descriptor::all()
                     .into_iter()
-                    .map(|theme| {
-                        DropdownItem::new(theme.name.to_string())
-                            .set_dispatches(Dispatches::one(Dispatch::SetTheme(theme.clone())))
+                    .enumerate()
+                    .map(|(index, theme)| {
+                        DropdownItem::new(theme.name().to_string())
+                            .set_rank(Some(Box::from([index].to_vec())))
```
wongjiahau, line 2079: Why rank by `index` instead of `theme.name()`?

jcowgar: Didn't know you could, will make the change.

src/themes/from_zed_theme.rs
```diff
@@ -1,210 +1,287 @@
...
-        .flat_map(|theme| -> anyhow::Result<Theme> {
+        .iter()
+        .find(|theme| theme.name == name)
+        .ok_or_else(|| anyhow::anyhow!("could not find theme '{}'", name))?
+        .clone()
+        .into())
+}
+
+impl From<ThemeContent> for Theme {
+    fn from(theme: ThemeContent) -> Self {
```
wongjiahau, line 110: I think we should avoid using `From`, because based on my experience, it makes tracking this part of the code difficult, IIRC running "Go to definition" on `into()` doesn't always lead to the specialized implementation.

We should keep this as a simple top-level function first, until one day we need a function to abstract over `T: From<ThemeContent>`.

wongjiahau: Also, unrelated to the above points, using a top-level function keeps the indentation of the body unchanged, thus greatly reducing space-wise diffs.

src/app.rs
wongjiahau: Oops sorry you're right, `set_rank` only takes usize.

wongjiahau (changes requested):
I'm good with the architectural changes overall, only minor style changes required.

src/app.rs
```diff
@@ -2071,11 +2075,15 @@ impl<T: Frontend> App<T> {
...
             PromptConfig {
                 on_enter: DispatchPrompt::Null,
-                items: crate::themes::themes()?
+                items: crate::themes::theme_descriptor::all()
                     .into_iter()
-                    .map(|theme| {
-                        DropdownItem::new(theme.name.to_string())
-                            .set_dispatches(Dispatches::one(Dispatch::SetTheme(theme.clone())))
+                    .enumerate()
+                    .map(|(index, theme)| {
```
wongjiahau, line 2081: Can we rename `theme` to `theme_descriptor` so that it's clearer?

wongjiahau (changes requested):
One last minor change

wongjiahau (approved):

PR #508, opened 2025-01-05
feat: differentiate between user and system buffers

jcowgar:
# The Problem

Currently a user can open a file and it appears in the buffer list as well as cycling through opened buffers. The system, however, can also open files that the user did not request. For example, the LSP server sends notifications can causes files to be opened and tracked for diagnostic reasons.

This causes confusion and makes it more difficult for a user to switch between files in the editor. When they open file A and file B; then try to cycle through buffers or switch buffers using `space b` and the system shows file A, file C, file D, file X, file Y, and file B. Not at all what they were expecting.

Further, when searching, you may enter a search term and it find results in 20 files. Immediately, those files are opened and added to the buffer list. So, a user may be working on file A, search for some reference material and then go back to editing file A. Later they open file B. Now, when going back to file A, those 20 files that were previously found via search (and maybe never even visited) are in the buffer list and cycle buffer.

# The Solution

This PR adds a `user` boolean value to the buffer. If the user opened the file, its `user` flag is set to `true`. If the system opens it, for searching, LSP diagnostics or something else, the `user` flag is set to `false`.

The Buffer Picker and Cycle Buffer only presents user opened files.

A system opened file can be converted into a user file if the user makes a change to a file and saves it. For example, via search `file_b.txt` is shown and the user edits that file and saves. `file_b.txt` now becomes a user file and is presented in the Buffer Picker and Cycle Buffer action.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -985,9 +985,14 @@ impl<T: Frontend> App<T> {
             return Ok(matching_editor);
         }
 
-        let buffer = Buffer::from_path(path, true)?;
+        let mut buffer = Buffer::from_path(path, true)?;
         let language = buffer.language();
         let content = buffer.content();
+
+        if option == OpenFileOption::Background {
+            buffer.set_user(false);
+        }
+
         let buffer = Rc::new(RefCell::new(buffer));
         let editor = SuggestiveEditor::from_buffer(buffer, SuggestiveEditorFilter::CurrentWord);
         let component_id = editor.id();
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -40,6 +40,7 @@ pub(crate) struct Buffer {
     decorations: Vec<Decoration>,
     selection_set_history: History<SelectionSet>,
     dirty: bool,
+    user: bool,
 }
 
 #[derive(Clone, PartialEq, Eq, Hash, Debug)]
@@ -74,7 +75,20 @@ impl Buffer {
             quickfix_list_items: Vec::new(),
             selection_set_history: History::new(),
             dirty: false,
+            user: true,
         }
+    }
+
+    pub(crate) fn set_user(&mut self, user: bool) {
+        self.user = user;
+    }
+
+    /// Was this buffer requested directly by the user for editing?
+    ///
+    /// Scenarios where it was not, but opened anyway include Global Search, LSP
+    /// diagnostic tracking.
+    pub(crate) fn user(&self) -> bool {
+        self.user
     }
 
     pub(crate) fn clear_quickfix_list_items(&mut self) {
@@ -660,6 +674,7 @@ impl Buffer {
         if let Some(path) = &self.path {
             path.write(&self.content())?;
             self.dirty = false;
+            self.user = true;
             Ok(Some(path.clone()))
         } else {
             log::info!("Buffer has no path");
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -219,6 +219,7 @@ impl Layout {
     pub(crate) fn get_opened_files(&self) -> Vec<CanonicalizedPath> {
         self.background_suggestive_editors
             .iter()
+            .filter(|(_, editor)| editor.borrow().editor().buffer().user())
             .map(|(path, _)| path.clone())
             .collect()
     }


src/buffer.rs
```diff
@@ -40,6 +40,7 @@ pub(crate) struct Buffer {
     decorations: Vec<Decoration>,
     selection_set_history: History<SelectionSet>,
     dirty: bool,
+    user: bool,
```
wongjiahau, line 43: Let's add a documentation comment to explain the necessity of this field. I don't want to solely rely on PR comments as that will be hard to trace if this field is moved to another struct, or if this struct is moved to another file.

wongjiahau, line 43: Let's make this clearer by:
1. Renaming `user` to `opened_by`
2. Use `enum BufferOpenedBy {Lsp, User}`

jcowgar: I did have a comment on it initially, but I then decided that it was private and never really used by someone outside of this file. Therefore, I added a comment to the `pub fn user()` function which describes what it is used for. That will show up in the editor where people are using it.

I'd be happy to add a comment to the field as well.

jcowgar: It can be more than Lsp or User, for example, search. We would then need to add to the `open_` functions an additional parameter to pass in this information, I think. What do you think?

wongjiahau: I think you can move the comments from the method to the field, and then document the method saying something like:
```rs
/// Refer `Buffer::opened_by`
```

src/buffer.rs
```diff
@@ -660,6 +673,7 @@ impl Buffer {
         if let Some(path) = &self.path {
             path.write(&self.content())?;
             self.dirty = false;
+            self.system_opened = false;
```
wongjiahau, line 676: Can you add a test case to test this path? Or more generically, the scenario `User saving a system-opened file will cause the file to appear in the buffer list`?

src/buffer.rs
```diff
@@ -660,6 +684,7 @@ impl Buffer {
         if let Some(path) = &self.path {
             path.write(&self.content())?;
             self.dirty = false;
+            self.owner = BufferOwner::User;
```
wongjiahau, line 687: If I did not misunderstand, this line is not tested, because this line is catered for this scenario, which is not tested: “Force Save a system-owned file that is unedited”

wongjiahau (changes requested):
I think we also need to explain in the docs regarding how buffer list work based on the logic in this PR.

src/buffer.rs
jcowgar: Actually, that line should be removed. Good catch. It becomes owned by a BufferOwner::User on edit now, not on save. I'll remove the line now.

wongjiahau: That works too, I guess force saving an unedited file is too unconventional after all.

src/app.rs
```diff
@@ -166,7 +160,7 @@ impl<T: Frontend> App<T> {
             if entry_path.as_ref().is_dir() {
                 self.layout.open_file_explorer();
             } else {
-                self.open_file(&entry_path, OpenFileOption::Focus)?;
+                self.open_file(&entry_path, BufferOwner::System, true)?;
```
wongjiahau, line 163: Why is this system owned?

src/app.rs
```diff
@@ -1873,7 +1852,7 @@ impl<T: Frontend> App<T> {
                     Direction::End => current_index + 1,
                 };
 
-                self.open_file(&files[next_index], OpenFileOption::Focus)?;
+                self.open_file(&files[next_index], BufferOwner::User, true)?;
```
wongjiahau, line 1855: The `BufferOwner::User` supplied here will be discarded.

src/app.rs
```diff
@@ -1886,7 +1865,7 @@ impl<T: Frontend> App<T> {
             Some(tagged_editor)
                 if Some(tagged_editor.clone()) != self.current_component().borrow().path() =>
             {
-                self.open_file(&tagged_editor, OpenFileOption::Focus)?;
+                self.open_file(&tagged_editor, BufferOwner::User, true)?;
```
wongjiahau, line 1868: Same here, we should create a method called `open_buffer` that only takes one parameter which is `CanonicalizedPath`, and `open_file` should call `open_buffer` if the requested path is already opened.

src/app.rs
```diff
@@ -2281,8 +2260,8 @@ pub(crate) enum Dispatch {
         scope: Scope,
         if_current_not_found: IfCurrentNotFound,
     },
-    OpenFile(CanonicalizedPath),
-    OpenFileFromPathBuf(PathBuf),
+    OpenFile(CanonicalizedPath, BufferOwner, bool),
+    OpenFileFromPathBuf(PathBuf, BufferOwner, bool),
```
wongjiahau, line 2264: Try to avoid tuples if there are more than 1 parameters, use key-value instead.

Luckily this can be done easily via Code Actions (`space c`) when using Rust Analyzer LSP.

src/layout.rs
```diff
@@ -80,8 +80,21 @@ impl Layout {
                 .background_suggestive_editors
                 .iter()
                 .skip_while(|(p, _)| p != &&path)
+                .skip_while(|(_, editor)| {
+                    editor.borrow().editor().buffer().owner() == BufferOwner::User
+                })
```
wongjiahau, line 85: Why?

src/layout.rs
```diff
@@ -80,8 +80,21 @@ impl Layout {
...
+                .or_else(|| {
+                    self.background_suggestive_editors
+                        .first()
+                        .and_then(|(path, editor)| {
+                            if editor.borrow().editor().buffer().owner() == BufferOwner::User {
+                                Some((path, editor))
+                            } else {
+                                None
+                            }
+                        })
```
wongjiahau, line 96: Why?

wongjiahau (changes requested):

src/layout.rs
jcowgar: Say you perform a search and it finds your term in 40 files. When closing the window, it will go to the 39th found file, then the 38th, 37th, etc... Close seems like you opened 40 files.

jcowgar: Oh, wait... hm. That should be `== BufferOwner::System`! Hm, my unit test is only opening two files. I see why it passed. I can fix.

src/layout.rs
jcowgar: Because if no User buffers were found, it was just taking the first one on the list after closing which could be a System buffer.

wongjiahau: If that's the case, I think:
```rs
                .skip_while(|(p, _)| p != &&path)
                .skip_while(|(_, editor)| {
                    editor.borrow().editor().buffer().owner() == BufferOwner::System
                })
                .nth(1)
                .or_else(|| {
                    self.background_suggestive_editors
                        .first()
                        .and_then(|(path, editor)| {
                            if editor.borrow().editor().buffer().owner() == BufferOwner::User {
                                Some((path, editor))
                            } else {
                                None
                            }
                        })
                 })
```
Can be reduced to:
```rs
                .skip_while(|(_, editor)| {
                    editor.borrow().editor().buffer().owner() == BufferOwner::System
                })
                .next() // Changed `.nth(1)` to `next`, as the former means the second element
```

If I'm understanding the code correctly, if `skip_while(...).next()` returns `None`, then the `or_else(...)` branch will also return `None`, thus `or_else(...)` is redundant.

wongjiahau (approved):
Looks good to me. Feel free to merge!

PR #509, opened 2025-01-05
feat: quick editor jumping

jcowgar:
The numbers 1-5 can be used to "tag" an editor with that number. When pressing the number:

1. If no editors are tagged with the number pressed, the current editor will be tagged with that number.
2. If the current editor is already tagged with the number pressed, the tag will be cleared.
3. If another editor is tagged with the number pressed, you will jump to that editor.

This provides a very quick and efficient way of managing the most important files to you for your given editing session. The idea was built from other editor's plugins such as Harpoon, Reach and Lasso for Neovim.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -736,6 +736,7 @@ impl<T: Frontend> App<T> {
             Dispatch::GoToPreviousFile => self.go_to_previous_file()?,
             Dispatch::GoToNextFile => self.go_to_next_file()?,
             Dispatch::CycleBuffer(direction) => self.cycle_buffer(direction)?,
+            Dispatch::JumpEditor(tag) => self.handle_jump_editor(tag)?,
             Dispatch::PushPromptHistory { key, line } => self.push_history_prompt(key, line),
             Dispatch::OpenThemePrompt => self.open_theme_prompt()?,
             Dispatch::SetLastNonContiguousSelectionMode(selection_mode) => self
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -1878,6 +1879,28 @@ impl<T: Frontend> App<T> {
...
+            Some(tag)
+        };
+
+        self.current_component()
+            .borrow_mut()
+            .editor_mut()
+            .set_tag(tag_to_set);
+
+        Ok(())
+    }
```
wongjiahau, line 1902: Let's make it clearer by showing the three different conditions syntactically:

```rs
fn handle_jump_editor(&mut self, tag: char) -> anyhow::Result<()> {
    let new_tag = match self.layout.find_editor_tagged(tag) {
        // Case 1: Found a non-current editor that has this tag
        //    - Jump to it
        Some(tagged_editor) if Some(tagged_editor.clone()) != self.current_component().borrow().path() => {
            return self.open_file(&tagged_editor, OpenFileOption::Focus)
        }
        // Case 2: Found current editor to have this tag
        //    - Remove current editor's tag
        Some(_) => {
            None
        }
        // Case 3: No editor found
        //    - Add new tag to current editor
        None => {
            Some(tag)
        }
    };
    
    self.current_component()
        .borrow_mut()
        .editor_mut()
        .set_tag(new_tag);
    Ok(())
}
```

src/layout.rs
```diff
@@ -310,6 +310,20 @@ impl Layout {
...
+        self.background_suggestive_editors
+            .iter()
+            .find_map(|(_, editor)| {
+                let suggestive_editor = editor.borrow();
+                let editor = suggestive_editor.editor();
+                if editor.tag() == Some(tag_char) {
+                    editor.path().clone()
+                } else {
+                    None
+                }
```
wongjiahau, line 323: Can be shortened using [`then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some).

wongjiahau (changes requested):

wongjiahau (approved):

PR #511, opened 2025-01-06
feat(language): if other methods fail, detect language from first line

jcowgar:
The filename does not always tell the story as to what a file is. Shell scripts are a prime example. you may have a file named: `get_weather` and in it the first line contains `#!/bin/bash`. Currently, this file will not be detected as a bash script and will gain none of the LSP/Tree-sitter goodness that Ki has to offer.

With this change, if the filename and file extension does not yield a result on the language to use, the first line of the file content will be examined looking for a common shebang or mode header line. Mode header lines considered are the popular vim syntax and another syntax started by (I think) Emacs and used in other editors as well.

If the first line of a file contains any of the following, the Bash language will be assigned in Ki.

- `#!/bin/bash`
- `# vim: ft=bash`
- `# mode: bash`

For the mode comment lines, the position of where `ft=` or `mode:` appear does not matter. Nor does any additional spacing around special characters.

### DIFF
--- a/shared/src/language.rs
+++ b/shared/src/language.rs
@@ -207,7 +207,7 @@ use regex::Regex;
 /// - `# mode: bash
 ///
 /// Spaces and other content on the line do not matter.
-pub fn from_content(content: &str) -> Option<Language> {
+pub fn from_content_directive(content: &str) -> Option<Language> {
     let first_line = content.lines().next()?;
 
     let re = Regex::new(r"(?:(?:^#!.*/)|(?:mode:)|(?:ft\s*=))\s*(\w+)").unwrap();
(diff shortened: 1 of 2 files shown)


shared/src/language.rs
```diff
@@ -192,6 +192,41 @@ pub(crate) fn from_filename(path: &CanonicalizedPath) -> Option<Language> {
...
+///
+/// For example, a file opened that has any of the following first lines will be
+/// detected as bash.
+///
+/// - `#!/bin/bash`
+/// - `# vim: ft=bash`
+/// - `# mode: bash
+///
+/// Spaces and other content on the line do not matter.
+pub fn from_content(content: &str) -> Option<Language> {
```
wongjiahau, line 210: Can we rename this to `from_content_directive`?

wongjiahau (changes requested):

wongjiahau (approved):

PR #544, opened 2025-02-16
add support for fish and unison programming langs

Diogenesoftoronto:
First pr to ki, said i would do this earlier

I want to say that I will be sort the langs later, just wanted to put this here before I forget to contribute.

- add treesitter for both
- add lsp for both (unison is unique)


wongjiahau:
Thanks!

PR #554, opened 2025-03-16
perf(buffer/undo,redo): store Edits instead of Diff patches

wongjiahau:
This is because diff patch is very slow when there are many cursors.


wongjiahau:
Hey @name-snrl I don't have plan to bring back undo tree so far, because to be frank I have not find the need to switch between undo branches.

Do you find undo tree necessary to your workflow?

wongjiahau:
> > Do you find undo tree necessary to your workflow?
> 
> Absolutely. I'd be glad to see undotree or similar functionality that provides a full persistent history of changes.
> 
> While I actively use fixup commits, which allows me to keep many intermediate states in reflog, there are often situations where I want to go back to states that are too small for new commit. Also, the contents of the reflog often look very messy, not to mention using `git fsck` to find staged files. On top of that, not all files are tracked by the VCS.
> At the moment, I still use neovim daily, and I like the functionality of telescope undotree.

I see, I'm open to adding undo tree back to Ki as long as it doesn't cause the same performance issue again.

> However, every day I try to use ki-editor more and more. I feel that what you’ve created is how editors should work in the age of tree-sitter and LSP.

Thanks for the encouragement!

> I’m very interested in the future of ki, so I also wanted to ask if you have a development roadmap?

There's no written roadmap at the moment, but one of the biggest goals is to make Ki embedable, as seen in https://github.com/ki-editor/ki-editor/pull/630.

What this means is that I want to export Ki motions to everywhere with a text area such as VS Code, Neovim, JetBrains RustRover, fish shell, macOS, etc, I do not wish Ki motions (which I think is the most efficient modal editor motions out there) to be trapped within the current Ki the terminal editor.

And I cannot thank @yves-biener enough for this spark, I have now been enjoying the best of both worlds by using Ki in VS Code, I get to have my cake and eat it too :) Before this, I had to sacrifice so many workflow enhancers (e.g., AI completion, GitLens) by restricting myself to work only within the terminal Ki.

Issue #558, opened 2025-03-24
System clipboard not working.


morfert:
I'm installing ki with the flake on NixOS and neither copy nor paste from the system clipboard are working. I am using the Ghostty terminal and nushell but I also tried Konsole terminal with bash and the clipboard commands did not work.
When i try to paste I get the following error:
 
"The clipboard contents were not available in the requested format or the clipboard is empty."

When I copy from ki and try to paste in a different app my system clipboard is empty.

wongjiahau:
Were you using `space c` (Copy to System Clipboard) and `space v` (Copy to System Clipboard)?

What window manager are you using?

morfert:
Yes, I used the space menu keybinds. I am using hyprland on wayland.

morfert:
That fixes it. I used the flake.nix to test and build but the flake could not find one of the new dependencies in its cargo-io flake so I had to update the flake.lock before it would build. Specifically the dependency `wl-clipboard-rs` could not be found until I updated the flake.lock.

morfert:
#559 is a commit I made to add optional support for wayland so that the clipboard will work.

morfert:
I am now using #559, this issue can be closed. Whether or not #559 gets merged can be discussed there.

haxfn:
> Can you try changing `arboard = "3.2.0"` in the root `Cargo.toml` of Ki to the following and rebuild Ki?
> 
> ```
> arboard = { version = "3.2", features = ["wayland-data-control"] }
> ```
> 
> ## Reference
> [1Password/arboard#98 (comment)](https://github.com/1Password/arboard/issues/98#issuecomment-1461267806)

This worked for me. Weirdly it installed arboard version 3.6 instead of 3.2

PR #559, opened 2025-03-25
added a `wayland` feature to the editor to allow the system clipboard…

morfert:
… to work on wayland

also added a secondary package to the flake that builds which builds with the `wayland` feature


morfert:
This commit adds an off by default feature, updates the cargo.lock and flake.lock and adds a second package to the flake which builds with the newly added feature enabled.

morfert:
This commit comes from trying to solve clipboard problems on wayland which I asked about #558 .

wongjiahau:
Thanks!

PR #560, opened 2025-03-25
Add support for language Ruby and Roc

jwoudenberg:
Hi there! I'm trying/learning Ki, loving it so far! There's so many cool ideas here!

I'm working with the languages Ruby and Roc quite a bit and added support for two languages I'm working with quite a bit: Roc and Ruby. Offering here to contribute them back in case it might be of benefit to others!

I tried to mirror the commit style I saw in other commits. Happy to make any changes, and of course no bad feelings if you don't feel like merging this!

### DIFF
--- a/shared/src/languages.rs
+++ b/shared/src/languages.rs
@@ -33,6 +33,7 @@ pub const LANGUAGES: &[&Language] = &[
     &nix(),
     &python(),
     &rescript(),
+    &roc(),
     &ruby(),
     &rust(),
     &sql(),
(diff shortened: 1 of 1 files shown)


jwoudenberg:
Oops, the Roc definition contained an error. Fixed now!

wongjiahau:
Thanks for the contribution  @jwoudenberg! Feel free to provide feedbacks, I'm more than happy to receive them :)

jwoudenberg:
I'm still wrapping my head Ki, and Vim muscle-memory is tripping me up 😄. The keyboard shortcut coherence means I've memorized a lot of them already, even if they haven't fully entered my muscle memory yet. I'm getting better though.

The set of batteries included with Ki is exactly what I want: project-wide search and a file explorer are the main things I've plugins for in Neovim.

I think the one thing I miss most at the moment from Neovim is delete not putting content on the clipboard. I think I'll get used to it though, after all copy+delete is only one keypress extra, and I haven't really incorporated 'Replace' in my workflow yet.

Is there a way to jump between marked selections?

I've briefly tried working with an LSP (I think the Nix one), but it was constantly showing error messages. That's probably the LSP in question and not Ki though. I try out LSPs every half year or so (previously with Neovim), and so far I've always found them to unstable for my taste.

wongjiahau:
Thanks for the feedback! 

>  I haven't really incorporated 'Replace' in my workflow yet.

`Replace` is useful if you want to replace another selection with the copied text.

For example:
1. Replacing the current line with the current selected word via `w c a x` (Word, Copy, Line, Replace).
2. Replacing an ancestor node with the current node via `d c i ... i x` (Syntax Node, Copy, Up [Expand selection], ... Up [Expand selection], Replace)


> Is there a way to jump between marked selections?

Yes, you can do it by using either `y` (Find Backward), `p` (Find Forward), or `n` (Find Global), then press `b` (Mark). Also, you can perform the same for all non-contiguous selections, such as Diagnostics, Git Hunk, etc.

jwoudenberg:
Thank you, this is super useful!

wongjiahau:
You're welcome!

PR #573, opened 2025-04-09
feat(doc): search bar using Lunr.js

wongjiahau:
image


wongjiahau:
Thanks for suggesting Lunr.js @rluiten, it is much easier than I thought!

rluiten:
Nice that someone solved it with lunr.js for docusaurus.
Looks pretty good deployed so far.

PR #578, opened 2025-04-11
feat(editor): first line should be selected by default

wongjiahau:
Since Line is the default selection mode.

### DIFF
--- a/src/components/editor.rs
+++ b/src/components/editor.rs
@@ -573,8 +573,10 @@ impl Editor {
     }
 
     pub(crate) fn from_buffer(buffer: Rc<RefCell<Buffer>>) -> Self {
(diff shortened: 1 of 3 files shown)


rluiten:
There's at least one more scenario where the line isnt selected.

I have this patch locally and seems to work for starting editor.
But once started i can hit `opt-k` to 1/2 page down and the line it lands on doesn't have the selected line cursor.

**I thought I typed this comment up somewhere already but can't find it :( so repeating it here**

wongjiahau:
> But once started i can hit opt-k to 1/2 page down and the line it lands on doesn't have the selected line cursor.

Ah ok, I know what you mean here, but this is a separate issue, let me create it.

PR #583, opened 2025-04-12
fix(help/keymap): handle multiline description widths

rluiten:
So the width used for cell width was using length of all lines.

This is not very important, if you think its not worth bothering with that's fine. 😄 

I had noticed there was a lot of padding when Help or Space menu was showing the shift and option keys.
My terminal width on my right monitor was wide enough for data without the extra padding and it was bugging me that it was wrapping. Took a while and some more practice reading Rust but I got there.

The original code used the length of all 3 lines of text concatenated for width not the longest line in a multiline scenario.

In addition for general String processing the original code used `len()` and not `chars().count()` so it doesnt handle multi byte UTF-8. But I left it with `len()` for now as original was len().

### DIFF
--- a/src/components/editor_keymap_printer.rs
+++ b/src/components/editor_keymap_printer.rs
@@ -143,7 +143,7 @@ impl KeymapPrintSection {
                         .unwrap_or_default() as u16
                 })
                 .max()
-                .unwrap_or_default() as u16;
+                .unwrap_or_default();
             ColumnConstraint::LowerBoundary(Width::Fixed(min_width.min(max_column_width)))
         };
 
--- a/src/components/editor_keymap_printer.rs
+++ b/src/components/editor_keymap_printer.rs
@@ -135,7 +135,13 @@ impl KeymapPrintSection {
                 .keys
                 .iter()
                 .filter_map(|row| row.get(column_index))
-                .map(|key| key.display(show_shift_alt_keys).len())
+                .map(|key| {
+                    key.display(show_shift_alt_keys)
+                        .lines()
+                        .map(|line| line.len())
+                        .max()
+                        .unwrap_or(0)
+                })
                 .max()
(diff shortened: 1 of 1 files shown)


rluiten:
I updated the code as I realised that UTF-8 chars are used in Ki key descriptions.
So this is actually better, it also means by less extraneous padding.

wongjiahau:
This is a nice change, just in case I misunderstood, you intended to reduce space wastage, as described by the following screenshots, right?

## Before

image

## After

image

wongjiahau:
Also, you might want to fix the lint error that causes the Github Actions pipeline to fail, as described at:

https://github.com/ki-editor/ki-editor/actions/runs/14427088211/job/40457388259#step:5:1003

rluiten:
Yeah those screen show demonstrate how much extra space there was in the past.|
Though github scales those images so its hard to the actual size change but you can see
the first image has a lot of padding left and right the second image basically has the 1 
character default comfy_table column padding.

Aside - intermediate data models. (this is not super important but a general design thing)
The way you put new lines in the bottom layer of key cell names, you are encoding data required
in the output to comfy_table in your core data model.
I think you'd be better off just returning a list of strings from that layer and then do whatever
in on in between layers as needed. Then the last point where you feed it to comfy_table is the only
thing that cares about newlines for multiple labels in a Cell so that's the only place you should new lines
So newlines are a data shape feature specifically for comfy_table and only comfy_table.

The displayed keys are much more consistently sized now.
I was frustrated with how wide my terminal need to be to not have garbled keymaps due to line wrapping.
The expanded keymaps with rest of keys now easily fit in most cases for me now.


Those 2 screen shots

rluiten:
Looking at the warning now thanks.

wongjiahau:
Thanks @rluiten!

PR #642, opened 2025-04-29
feat(editor): vertical movement sticky column

wongjiahau:


rluiten:
@wongjiahau I've got to use the sticky column feature now so far its great.

PR #643, opened 2025-04-30
feat(app): double status line

wongjiahau:


wongjiahau:
@rluiten if you're free, please let me know what you think about this, I've added one more row of status line to house more information like process ID, `line:column` etc.

image

rluiten:
I hate to be "that guy" I tend to value my vertical screen real estate a lot.
I was already thinking about getting rid of the filenames in the kind of tab bar across the top :).
Or thinking how to minimise on screen all the time info and make it avail some other way maybe.

What does the "File" mean on the right upper line?

wongjiahau:
I get what you mean, two status lines  actually only make sense when I'm using a vertical monitor.

The "File" on the right actually means the last action, it will change as you execute other actions like movement or edits.

rluiten:
I've been running 2 monitors of 16:10 aspect ratio monitors for 20 years.
The left being landscape orientation the right being portrait.
The left being centered vertically against the right :)

vdawg-git:
Something like this could be useful as a config option.
Some people like to have a lot of information visible, others prefer minimalism. 
And both are valid

wongjiahau:
> Ki is getting continuously better. Ki I believe already pretty suckless.

🙇 

> One of the concepts of ki is: "Speed and efficiency for experienced users"[1](#user-content-fn-1-65bc9ba85a9bc4b260ce9c3c5c72707d), with that in mind I have some shareable opinions:
> 
> 1. **Directory path can be shortened as in**
>     image
>    This would save a lot of space in the status bar in ki, which currently would take more space if the path is deeper:
>     image

This is a good suggestion, perhaps if the user wants to know the pwd, we just have to implement a command for that.

> 2. **Remove "almost-static" stuff from the status bar: Current Keyboard Layout**
>    For example there is very little informational value in showing the current layout as one it is only infrequently going to change and second users that indeed change current layout frequently will change it in other places system wide not just the editor so they must have some other indicators outside of editor for them. For me I have an indicator in my status bar.

TBH, I thought I needed it (because I constantly swap between Dvorak-IU and Qwerty), but most of the time, I don't have to look at the status bar to know that the current keyboard layout is not matching my current keyboard, because the command goes haywire if they doesn't match.

So, yup, let's remove it.

> 3. **Help(alt+m) should only be shown on the scratch buffer**

Good idea. But the prerequisite for this is that Ki needs to allow the concept of scratch buffer, it's not currently done yet, because I don't find it necessary, but I'm not against it either.

haxfn:
I am now of the opinion that two status bars are alright:

- vim used to have command input line
- emacs also has a command input line

haxfn:
many of the hyper optimisations (like fish-like working directory path, shortening/trimming git branch) are skipped if we use two status bars. There is a lot of flexibility that comes with two status bars.

haxfn:
opening explorer opens the project folder and not where this file is.

haxfn:
Yes. We should settle on what all things to put in the status bars. Minimal yet sufficiently informational is best.

> fast for experience users

should be the focus, I think

Issue #666, opened 2025-06-22
Incompatible Tree-Sitter ABI version


name-snrl:
I ran into the same issue that's described here https://github.com/zed-industries/zed/issues/24632

```
[00:00:00.025] (7f6daa25e000) INFO   ProcessCommand::spawn: Failed to locate "bash-language-server"
[00:00:00.176] (7f6da97156c0) INFO   Loading from /home/name_snrl/.config/ki/grammars/sources/bash/queries/highlights.scm
[00:00:00.178] (7f6da97156c0) INFO   syntax_highlight_error = QueryError {
    row: 0,
    column: 0,
    offset: 0,
    message: "Incompatible language version 15. Expected minimum 13, maximum 14",
    kind: Language,
}
```

PR #667, opened 2025-06-22
fix(Cargo.toml): bump tree-sitter versions

tsalkenov:
should fix bash grammar ABI incompatability [here](https://github.com/ki-editor/ki-editor/issues/666)


wongjiahau:
Thanks!

PR #700, opened 2025-07-24
fix: keymap display for smaller windows

mjakobsche:
Hi, 
I'm using 90-char width terminal, and the keymap legend layout gets destroyed becoming unreadable. I came up with a simple fix - enable dynamic content arrangement and refresh the table size after creation:
 
Image

Its still not perfect - for some smaller sizes that problem appears again like before:

image

 Perhaps there is a better way to resolve it?


wongjiahau:
Thanks! Can you just `just fmt` to format the files?

> Perhaps there is a better way to resolve it?

I think we might have to do a Stack layout in this case, which is to stack the Left section above the Right section:

image

While we are searching for better solutions, have you tried zooming out your terminal temporarily when this happens?

wongjiahau:
> Maybe the cleanest approach would be to display table only if it can fit - and else just show a subtle message like „terminal window is too small to display kemymap legend”.

> What do you think about that?


I think this is a good idea, let's do it!

PR #701, opened 2025-07-24
feat(languages/ruby): add fallback, extensions and filenames for ruby

mjakobsche:
this extends ruby language support, by providing:
- language fallback
- more file extensions
- and some popular ruby-specific file names


wongjiahau:
Thanks @mjakobsche! Can you just `just fmt` to format the files?

PR #718, opened 2025-08-05
add support for xaml & axaml files

pragun:
XAML files are used by WPF, and AXAML files are used by Avalonia. They're XML files sytactically so XML tree sitter works fine


wongjiahau:
Thanks @pragun!

PR #751, opened 2025-08-19
fix(docs/intro) add assertion and change `delete` keystroke

haxfn:
image

In [Docs Intro](https://ki-editor.github.io/ki-editor/docs/introduction/), the last key be something like 'v' (to delete) instead of 'h' (entering insert mode).

### DIFF
--- a/src/recipes.rs
+++ b/src/recipes.rs
@@ -1419,11 +1419,37 @@ pub(crate) fn run(path: Option<CanonicalizedPath>) -> anyhow::Result<()> {
                     .trim(),
                     file_extension: "rs",
                     prepare_events: &[],
-                    events: keys!("q p r i n t enter r r d h"),
-                    expectations: Box::new([]),
+                    events: keys!("q p r i n t enter r r d v"),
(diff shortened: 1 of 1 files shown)


src/recipes.rs
```diff
@@ -1419,11 +1419,37 @@ pub(crate) fn run(path: Option<CanonicalizedPath>) -> anyhow::Result<()> {
...
+    });
+
+    app.run(path)
+        .map_err(|error| anyhow::anyhow!("screen.run {:?}", error))?;
+}"#
+                    )]),
                     terminal_height: None,
                     similar_vim_combos: &[],
-                    only: false,
+                    only: true,
```
wongjiahau, line 1452: You need to change this to false otherwise the CI will fail.

wongjiahau (changes requested):

src/recipes.rs
haxfn: Ah, ok

haxfn: Done

wongjiahau (approved):

wongjiahau:
Thanks @haxfn!

PR #763, opened 2025-08-23
feat(docs/interactive-tutorial): copy button for current instance content

### LINKED ISSUE #761: doc: Suggestion for copy buttons for interactive examples
The interactive examples is really a great thing. It makes it easier to digest the keybinds one key-stroke at a time.

Once in a while a nice and long example shows up. I naturally want to copy these examples and play with them in my own instance(terminal).

For essence, consider the docs site example on deleting multiple `println!()` statements. It is sluggish to copy the long example.

How about we provide a button to copy:

1. Copy the current contents of the instance at any intermediate keystroke. Or
2. Copy only the starting contents of the example. Or
3. Copy both the starting and end contents.

The third point here is close at heart to vim-golf kind of thing but for `ki`. But maybe another issue for that.

haxfn:
Fixes #761 

TODO:

1. - [x] Update Json producer to include per step editor buffer content
        - [x] Modify type signatures of helper functions
        - [x] Capture buffer content in intermediate `callback` closure

2. - [x] Correct buffer content:
        - [x] Refactor `(Option<String>, Option<String>)` to 
            ```rust
            struct TestOutput {
                term_output: Option<String>,
                buffer_contents: HashMap<String, String>,
            }
            ```
        - [x] Expose ~~`Vector<Editor>`~~ `BufferContentsMap` for `App` under `Layout::background_suggestive_editors()`

3. - [x] Update Json consumer to copy immediate buffer contents to clipboard (multiple copy buttons, one for each buffer)
4. - [-] Expose last active buffer name (Skipped and reported in a new issue)
       - [-] Track and save last active buffer in App state (?)
6. - [-] Fix Json consumer copy behaviour (single copy button for last active buffer) (Skipped in preference of sorting based on the file name)

Future:

1. issue #766 
2. Instead of sorting based on file-name, using any meanigful ordering in the indexmap `Layout::background_suggestive_editors: IndexMap<CanonicalizedPath, Rc<Ref<Editor>>>`. This might be marginally more intuitive for the copy button order.

### DIFF
--- a/docs/src/components/Tutorial.tsx
+++ b/docs/src/components/Tutorial.tsx
@@ -70,6 +70,9 @@ const Recipe = (props: { recipe: Recipe }) => {
     const step = props.recipe.steps[stepIndex];
     instance?.write(step.term_output);
   }, [ref, instance, stepIndex]);
+  const buffer_contents_entries = Object.entries(props.recipe.steps[stepIndex].buffer_contents_map)
+    .sort((a, b) => (a[0].localeCompare(b[0])));
+
 
   return (
     <div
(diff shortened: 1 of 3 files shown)


haxfn:
image

Currently: multiple copy buttons, one for each buffer in the list of buffers.

wongjiahau:
Thank you!

PR #774, opened 2025-08-30
feat(ci): assertion for broken keymaps

### LINKED ISSUE #773: feat(ci): assertion for broken keymaps/tutorials
At this stage broken docs maybe off putting and confusing to new comers.

There should be a way to catch broken docs components like keymap-drawings or interactive tutorials on compile/build time. This would be a necessary annoyance in CI. This would serve to avoid broken docs. 

Two examples of broken docs are:

Image

Image

Hence, it is clear that this issue is the root cause of issues/#767 and duplicate/related issues mentioned therein.

haxfn:
Fixes #773 

Traverses AST of md/mdx files and checks at ci build time if the arguments to `<TutorialFallback filename="argument" />` is a valid static resource somewhere inside `docs/static/`

image

TODO:

- [ ] Use generators in `validate.tsx` to keep code mutation free (note-to-self: why?)


wongjiahau:
To fix the `tsx` not found issue, you can update flake.nix to include it

wongjiahau:
Nice job @haxfn! Can you add unit tests for your script? The test cases shouldn't involve file IO, they should just receive a string and output either success or error.

And of course the unit test should be included in CI too.

docs/validate.ts
```diff
@@ -0,0 +1,74 @@
...
+    const nodeType = node.type;
+    const isFlowElement = nodeType === 'mdxJsxFlowElement';
+    const isJsxTextElement = nodeType === 'mdxJsxTextElement'
+    const isNameTutorialFallback = node.name === 'TutorialFallback';
+    if ((isFlowElement || isJsxTextElement) && (isNameTutorialFallback)) {
+      const fileNameAttr = node.attributes?.find(
+        attr => attr.type === 'mdxJsxAttribute' && attr.name === 'filename'
+      );
+      if (fileNameAttr && fileNameAttr.value) {
+        argFileNames.push(fileNameAttr.value);
```
wongjiahau, line 23: This is optional, but I think we might be able to avoid mutations by using Rx.js:


```ts

const { fromMarkdown } = require('mdast-util-from-markdown');
const { mdxFromMarkdown } = require('mdast-util-mdx');
const { mdx } = require('micromark-extension-mdx');
const { Observable } = require('rxjs');
const { filter, map, toArray } = require('rxjs/operators');

// Create an observable that traverses the tree without mutations
function createNodeStream(tree) {
  return new Observable(subscriber => {
    function traverse(node) {
      subscriber.next(node);
      if (node.children) {
        node.children.forEach(traverse);
      }
    }
    traverse(tree);
    subscriber.complete();
  });
}

export function extractArgumentFileNames(mdxContent) {
  const tree = fromMarkdown(mdxContent, {
    extensions: [mdx()],
    mdastExtensions: [mdxFromMarkdown()]
  });

  return createNodeStream(tree)
    .pipe(
      filter(node => {
        const nodeType = node.type;
        const isFlowElement = nodeType === 'mdxJsxFlowElement';
        const isJsxTextElement = nodeType === 'mdxJsxTextElement';
        const isNameTutorialFallback = node.name === 'TutorialFallback';
        return (isFlowElement || isJsxTextElement) && isNameTutorialFallback;
      }),
      map(node => {
        const fileNameAttr = node.attributes?.find(
          attr => attr.type === 'mdxJsxAttribute' && attr.name === 'filename'
        );
        return fileNameAttr?.value;
      }),
      filter(filename => filename !== undefined),
      toArray()
    );
}

```

Note: I'm not sure if this works though, because it's generated by Claude.

wongjiahau (changes requested):

docs/validate.ts
```diff
@@ -0,0 +1,90 @@
...
+    .map(extractFilename)
+}
+
+function validateResourceAccess(mdxContent: String, validFilenames: Array<String>) {
+  const argFilenames = extractArgumentFileNames(mdxContent);
+  const validResources = argFilenames.map(argFileName =>
+    validFilenames.includes(argFileName)
+  );
+
+  let hasInvalidResource = false;
```
wongjiahau, line 52: Mutation can be removed by using the existential method (every/some in JS):

```ts
const hasInvalidResource = validResources.some(isValid => !isValid)
```

Theoretically speaking, iterating `validResources` twice (the second time for logging) is less performant than performing both operations in the same iteration (like the current, where it logs and computes `hasInvalidResource` in a single iteration), but pragmatically speaking, the performance improvement is negligible, in this codebase, code maintainability (both readability and extendability) is prioritized over trivial performance improvements.

docs/validate.ts
```diff
@@ -0,0 +1,92 @@
...
+    .map((testFilePath: string) => fs.readFileSync(testFilePath, 'utf8'))
+    .map((mdxContent: Value) => validateResourceAccess(mdxContent, validResourceFilenames));
+
+  validAccesses.forEach((validAccess: boolean, index: number) => { 
+    if (!validAccess) {
+      throw new Error(`Invalid static resource access in ${mdxFilePaths[index]}`)
+    }
+  });
+
+  if (validAccesses.every(Boolean)) { console.log("\t ALL STATIC RESOURCE ACCESSES in md/mdx WERE VALID!") }
```
wongjiahau, line 91: The condition `if (validAccesses.every(Boolean))` should be unnecessary.

docs/validate.ts
```diff
@@ -0,0 +1,92 @@
...
+  const mdxFilePaths = glob.sync('docs/**/*.{md,mdx}');
+  let validAccesses = mdxFilePaths
+    .map((testFilePath: string) => fs.readFileSync(testFilePath, 'utf8'))
+    .map((mdxContent: Value) => validateResourceAccess(mdxContent, validResourceFilenames));
+
+  validAccesses.forEach((validAccess: boolean, index: number) => { 
+    if (!validAccess) {
+      throw new Error(`Invalid static resource access in ${mdxFilePaths[index]}`)
+    }
+  });
```
wongjiahau, line 89: If I'm understanding correctly (please correct me if I'm wrong), when you mentioned better error logging, you were saying that the previous code shows all invalid accesses, unlike this newer code, which shows only the first invalid access.

This can be solved by using `map` to combine all invalid accesses into a single message before logging.


For example:
```ts
const erroneousFilePaths = validAccesses
    .flatMap((validAccess, index) => validAccess ? [] : [mdxFilePaths[index]])

if(errorneousFilePaths.length > 0) {
    throw new Error(errorneousFilePaths.join("\n"))
}
```

wongjiahau: Wait, I think the code can be further improved.

docs/validate.ts
```diff
@@ -0,0 +1,92 @@
...
+  );
+
+  validResources.forEach((isValid, index) => {
+    if (!isValid) {
+      console.log(`\tERROR: NON-EXISTENT STATIC RESOURCE:\t "${argFilenames[index]}"`);
+    }
+  });
+
+  return validResources.every(isValid => isValid);
+}
```
wongjiahau, line 63: In FP, we will try to keep as many functions as pure as possible, and only do IO at the program boundaries.

So this function shouldn't do a console.log and return a boolean, instead, it should return a list of usage of non-existent resources.

Keeping a function pure makes it better for testing and debugging.

wongjiahau (approved):

PR #860, opened 2025-10-05
Intellij plugin

Exidex:
This is a draft PR to develop in the open (in case I do not have will power to finish this, hehe). The development setup is done, IPC with Ki is working and ping action is implemented. I will update the PR as I go


angelozerr:
@Exidex it seems you have implemented your own ls support but are you aware that JetBrains provide an lsp support and also https://github.com/redhat-developer/lsp4ij

angelozerr:
@Exidex sorry for the confusion, but I wonder the benfit to use the bridge from ki because I find there are a lot of code. With a pure lsp you should have just few code. But I dont know ki, so perhaps I misse something.

PR #1036, opened 2025-12-05
feat(languages): add Julia

haxfn:
- https://github.com/julia-vscode/LanguageServer.jl


wongjiahau:
@haxfn there's one test case that is very flaky, it might even fail after 4 retries. For now simply re-running the CI should fix the issue.

haxfn:
I did rerun atleast 5 times before a Merge commit passed the CI

PR #1037, opened 2025-12-06
doc: fix: d selects largest node in QWERTY

rofrol:


haxfn:
I'll combine all the syntax-recipes related prs into one branch [rofrol-syntax-recipe-patches](https://github.com/ki-editor/ki-editor/tree/rofrol-syntax-recipe-patches)

PR #1041, opened 2025-12-06
doc: with D, you need to press j to select parent

rofrol:


haxfn:
Thanks for the catch. I believe, it should be `i` instead of `j` or `k`.

PR #1085, opened 2025-12-26
Don't extend zero-length ranges

### LINKED ISSUE #1079: Deleting the last character with a zero-length range panics
Running Delete One (vv in QWERTY) on a zero-length selection at the end of the file causes ki to panic. [stack trace](https://github.com/user-attachments/files/24335096/log.txt).

The relevant code is [this if statement, which extends the range](https://github.com/ki-editor/ki-editor/blob/a59dac72198a8063b7168c80ebd1a6971c29d653/src/components/editor.rs#L1669).

Riolku:
Fixes #1079

### DIFF
--- a/src/components/editor.rs
+++ b/src/components/editor.rs
@@ -1666,7 +1666,9 @@ impl Editor {
 
                     // Ensure the delete range is at least one character long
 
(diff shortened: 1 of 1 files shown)


Riolku:
Manually confirmed this solution works

PR #1098, opened 2025-12-29
feat(themes): add a runtime dir with default themes

haxfn:


haxfn:
Closed as pull/#1114 embeds default themes into the binary

PR #1105, opened 2025-12-29
feat(themes): Embed zed themes into the binary

### LINKED ISSUE #1097: feat(themes): add a runtime dir with default themes

Riolku:
Fixes #1097.

Embedding the themes as constants proved to be too frustrating to implement. Instead, the JSON5 documents are embedded into the binary and parsed at startup.


Riolku:
Going to outright close this. The better approach is to convert zed themes to Ki themes before build time. Zed themes are too big to embed outright.

PR #1107, opened 2025-12-30
refactor(deps): Use std instead of once_cell

Riolku:
std has good replacements for once_cell since 1.80


Riolku:
stacked on #1106

PR #1114, opened 2025-12-31
refactor(themes): embed compressed zed themes into the binary

Riolku:
Instead of downloading zed themes and caching them, we can directly embed them into the binary. This change also refactors ThemeDescriptors a bit, which is a bit more ergonomic and helps prepare better for custom theme support.

Why not convert to a Ki theme, then embed? Because this requires some sort of build script or pre-build step that converts, and doing that right now is arduous at best. Besides, we need a zed theme converter in the binary anyway, since we've more or less decided that custom theme schema will simply be the Zed theme schema.

Why compress the Zed themes? Because they are 10x the size (around 700K) otherwise, whereas presently they are ~60K.

Why miniz_oxide? It's a tiny library and an extremely popular crate: at present the rust standard library even depends on it.

Last but not least, this change alters the zed theme parsing from serde_json5 to serde_json_lenient. This is because zed actually uses the latter.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -14,7 +14,8 @@
             "src/protocol/types.ts",
             "target",
             "docs/static",
-            "docs/.docusaurus"
+            "docs/.docusaurus",
+            "themes"
         ]
     },
     "formatter": {
--- a/zed_theme/src/lib.rs
+++ b/zed_theme/src/lib.rs
@@ -1,3 +1,36 @@
-pub mod zed_theme_schema;
+mod zed_theme_schema;
 
 pub use zed_theme_schema::*;
+
+use std::collections::HashMap;
+
+const COMPILED_THEME_BYTES: &[u8] =
+    include_bytes!(concat!(env!("OUT_DIR"), "/compiled_themes.bin"));
+
+pub fn get_zed_themes() -> HashMap<String, ThemeContent> {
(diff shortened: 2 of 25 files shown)


biome.json
```diff
@@ -14,7 +14,8 @@
             "src/protocol/types.ts",
             "target",
             "docs/static",
-            "docs/.docusaurus"
+            "docs/.docusaurus",
+            "themes"
```
wongjiahau: I think the theme JSON files should be formatted so that in the future if we update the theme files, we won't have to commit formatting-only changes.

wongjiahau (changes requested):

zed_theme/src/lib.rs
```diff
@@ -1,3 +1,34 @@
...
+        serde_json_lenient::from_slice(&decompressed_themes)
+            .expect("Compiled themes aren't valid lenient JSON?");
+
+    let mut themes = HashMap::new();
+    for theme_family in theme_families {
+        for theme in theme_family.themes {
+            themes.insert(theme.name.clone(), theme);
+        }
+    }
+    themes
```
wongjiahau: Same here, this could be done using `.into_iter()` and `collect()`.

Riolku: Right, oop

wongjiahau (approved):

wongjiahau (approved):

PR #1129, opened 2026-01-01
tidy(deps): depend on crossbeam_channel instead of all of crossbeam

Riolku:


wongjiahau:
I wonder if we can just remove `crossbeam` by using `std::mspc::channel` as explained in https://google.github.io/comprehensive-rust/concurrency/channels/unbounded.html

Riolku:
Hm. You're correct, Ill change this. For some reason I thought we had multiple consumers

wongjiahau:
Tested this PR, Global Replace All still works. Thanks!

PR #1130, opened 2026-01-01
fix(build): make nvim-treesitter build.rs more robust

Riolku:
Ensure only whitelisted languages are allowed to have missing queries. This would've caught the painful CI error caused by forgetting to init submodules.


wongjiahau:
Thanks!

PR #1132, opened 2026-01-02
refactor(buffer): make get_edit_transaction functional style

Riolku:
A key observation here is that `.new_slices()` gives us access to a slice of lines, which is extremely useful for conveniently accessing the same representation of lines that similar has.

Also, this function cannot fail except in the event of a programming error.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -18,6 +18,7 @@ use crate::{
     syntax_highlight::{HighlightedSpan, HighlightedSpans},
     utils::find_previous,
 };
+use anyhow::Context as _;
 use itertools::Itertools;
 use regex::Regex;
 use ropey::Rope;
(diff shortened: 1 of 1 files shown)


wongjiahau (changes requested):
Based on my experience with coding Ki, I realise that anything that can fail, will fail.

There are many causes of crashes, even if the algorithm seems correct, one of them being garbage inputs.

That's why I gradually removed all potentially panicking function calls in the codebase, because they cause Ki to crash when I was certain it wouldn't happen lol.

To keep the code in a functional style, you can use either `try_collect` or `collect::<anyhow::Result<_>>()`

wongjiahau:
Hey! Sorry I was driving just now, regarding the potential panics, garbage input is just one of many possible unknown causes.

I agree with you that we should find out why it failed, but where I disagree is that we should not crash the app when it fails, especially for an editor.

And the current infrastructure will actually show the errors to the user (via GlobalInfo), without crashig the app, so yes I’ll insist on removing the panicking code.

Riolku:
Forgot to update commit message

PR #1134, opened 2026-01-03
feat(ci): nightly builds

Riolku:


Riolku:
Here is [a "passing" run](https://github.com/ki-editor/ki-editor/actions/runs/20670021422/job/59349509755?pr=1134). It can't possibly pass for real without a token with write permissions, which is only available on master.

PR #1135, opened 2026-01-03
fix(ci): rename nightly artifacts

Riolku:
Even if artifacts contain a single file, they still end up as an archive, which when downloaded, extract into a folder. I suspect that when uploaded to a release, they are not compressed, so we should rename before uploading.


Riolku:
[Build I'm happy with](https://github.com/ki-editor/ki-editor/actions/runs/20672063995?pr=1135)

PR #1136, opened 2026-01-03
refactor(deps): make wayland truly optional

Riolku:
The way the feature is set up right now, wayland is always compiled. This now makes it so wayland is enabled by default, but can be disabled.

Also, tempfile is not a regular dep, it's just a dev dependency.

Also, ki-vscode Cargo.toml is not used.


wongjiahau:
Thanks!

PR #1139, opened 2026-01-03
feat(ci): add tagged release workflow

Riolku:
By refactoring the CI workflows into a release-common and separate nightly & tagged release workflows, we can ensure nightly and tagged releases have the same build process.

Then, it's simple enough to make a new tagged release workflow that triggers whenever we push to a release tag.


wongjiahau:
Thanks!

PR #1144, opened 2026-01-04
refactor(build): vendor highlight queries

### LINKED ISSUE #1138: git subtree instead of git submodule?
Hi there! I noticed a `.gitmodule` was recently added containing a submodule, which unfortunately broke `nix build` (totally understandable, I don't think many people are relying on it!).

I tried for a bit to get `nix build` to work with submodules, but without success. It looks like there's an experimental nix feature in the pipeline that might make it easier in the future. I could wait for that.

There's a sort-of-alternative to git submodules called git subtrees which would also address this problem, which is also built into git. I wanted to check if there's some appetitie for trying it out, given git submodules have been known to cause a mild headache here and there outside of nix builds :). For instance, when using git subtrees there's no need for contributors of Ki to run `git submodule init` before they can build. This article gives a summary of the pros and cons:
https://www.atlassian.com/git/tutorials/git-subtree

If there's an interest I can make a PR switching us over.

Riolku:
Fixes #1138. The script in here does a good job of making updating less painful, and a `.ignore` file added to the folder makes global search in ki ignore the directory. Other than that, various things can now be cleaned up because there's no need to have submodule support baked into other parts of the application.

The directory is 1.7M, but the toplevel src/ directory is 2.3M. There's some room for optimization here, potentially by pruning to only the languages we need, but this adds a lot of complexity and trims it down to around 600K, which doesn't seem especially worth it.


Riolku:
> I think there are too many files committed and Github's code review UI is hanging because of that.
> 
> To simplify code review in the future regarding highlight queries update in the future, would it possible if we commit one a single compressed Zip file of the highlight queries, and unpack it during compilation (which I guess wouldn't affect the build time too much)?

I object to compressing the query files for a different reason: if we did that, every update effectively adds 1.7M to a `git clone`. Meanwhile, if we do it this way, size is added in small chunks for every update, since git can efficiently compute diffs.

wongjiahau:
That's a good objection. Just another question, should we have an automated cron job to update the highlight queries every now and then?

wongjiahau:
@Riolku how about once a week? What do you think?

wongjiahau:
Ok monthly it is!

PR #1147, opened 2026-01-04
feat(ci): more platforms

Riolku:
Support musl linux, windows ARM, and macOS intel.


Riolku:
It works: https://github.com/ki-editor/ki-editor/actions/runs/20688021261

PR #1148, opened 2026-01-04
doc: update README with installation instructions

Riolku:


wongjiahau:
Question, can we update the installation page in the docs and link it in the README instead? 

Or do you think it's better to have them directly accessible in README?

Riolku:
Hm, good point. I'll think more on this and revisit tomorrow.

Riolku:
closed in favour of #1157

PR #1170, opened 2026-01-06
refactor(organization): move ki to ki-bin subfolder

Riolku:
This is largely in preparation for a crate split. It also makes it more clear what Cargo.toml keys pertain to the workspace versus which pertain to the toplevel package.

The original behaviour of `cargo build` is retained with `default-members`.


Riolku:
The test structure relies heavily on the CWD, which causes lots of breakage. Closing for now, will revisit.

PR #1171, opened 2026-01-06
feat(build): make ki use lib.rs

Riolku:
This allows integration tests (a future addition) to use code in ki, it allows other (future) crates to import code in ki, and it's explicitly called out as a [pattern in the rust book][1].

[1]: https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates


Riolku:
code changes are just `mv src/main.rs src/lib.rs`

PR #1203, opened 2026-01-10
feat(ci): add cargo-workspace-unused-pub workflow

Riolku:
cargo-workspace-unused-pub has false positives even on our current workspace. However, having it create an issue every month would alert us to this kind of dead code in a fairly low-noise manner.

Also, I've trimmed some it has found already.


wongjiahau:
Thanks @Riolku!

PR #1204, opened 2026-01-10
tidy(ci): remove linux glibc builds

Riolku:
musl builds are statically linked, and so work on all linux platforms. Shipping only a musl build is also the approach taken by jujutsu.


wongjiahau:
Thanks!

PR #1232, opened 2026-01-13
keymap: Mark MoL

haxfn:
Moves alt+e to ek and alt+<<>> to e<<>>

additionally introduces alt+|<>| for first/last marked file

also turns off "warping" to first file when alt+>> on last file
and similary "warping" to last file when alt+<< on first file

### DIFF
--- a/src/test_app.rs
+++ b/src/test_app.rs
@@ -3670,7 +3670,7 @@ fn navigating_to_marked_file_that_is_deleted_should_not_cause_error() -> anyhow:
             )),
             Expect(CurrentPath(s.gitignore())),
             App(DeletePaths(NonEmpty::new(s.main_rs()))),
-            App(CycleMarkedFile(Movement::Left)),
+            App(CycleMarkedFile(Movement::Last)),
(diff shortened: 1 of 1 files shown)


src/components/editor_keymap_legend.rs
```diff
@@ -1733,6 +1733,28 @@ pub fn cut_keymaps(context: &Context) -> Keymaps {
...
+            (Meaning::Prev_, Movement::Previous),
+            (Meaning::Next_, Movement::Next),
+            (Meaning::First, Movement::First),
+            (Meaning::Last_, Movement::Last),
+        ]
+        .into_iter()
+        .map(|(meaning, movement)| {
+            Keymap::new(
+                context.keyboard_layout_kind().get_key(&meaning),
+                movement.format_action("Navigate File"),
```
wongjiahau: I think naming it `Marked File` should be clear enough, because the movement will add have added the arrows, for example, Left = `<< Marked File`, Last = `Marked File >|`.

PR #1237, opened 2026-01-15
keymap: Mark MoL fine movement should navigate open buffer files

haxfn:
With this change, Open buffer files are insignificant and marked files are significant.

- `>` `<` Fine movements navigate Opened Files and Marked Files
- `>>` `<<` Coarse movements navigate Marked Files only
- `>|` `|<` movements navigate Last / First Marked File

### DIFF
--- a/docs/blog/2026-01-10-evolution-of-delete.md
+++ b/docs/blog/2026-01-10-evolution-of-delete.md
@@ -238,7 +238,7 @@ For example, to delete two selections:
 
 Since most deletions involve only one selection, the menu approach is more efficient for the common case while keeping the Delete Submode available for users who prefer chaining deletions.
 
-## Version 6: Momentary Layer
+## Version 6: Momentary Layer (MoL)
 
 [Momentary Layer](https://zmk.dev/docs/keymaps/behaviors/layers#momentary-layer) combines the best aspects of both the Delete Menu and Delete Submode approaches: it's as efficient as the Delete Menu for single deletions while being as repeatable as the Delete Submode for chained operations.
 


src/app.rs
```diff
@@ -2197,32 +2198,178 @@ impl<T: Frontend> App<T> {
...
+                marked_file_paths.iter().for_each(|path| {
+                    if !merged_file_paths_map.contains_key(*path) {
+                        merged_file_paths_map.insert((*path).clone(), true);
+                    }
+                });
+                buffer_file_paths.iter().for_each(|path| {
+                    if !merged_file_paths_map.contains_key(path) {
+                        merged_file_paths_map.insert((*path).clone(), false);
+                    }
+                });
```
wongjiahau: You can make use of the [`chain`](https://doc.rust-lang.org/beta/std/iter/trait.Iterator.html#method.chain) method to join two iterators together without needing explicit loops:



```suggestion
marked_file_paths
        .iter()
        .map(|path| ((*path).clone(), true))
        .chain(
            buffer_file_paths
                .iter()
                .map(|path| ((*path).clone(), false))
        )
        .collect::<IndexMap<CanonicalizedPath, bool>>()
```

src/app.rs
```diff
@@ -2197,32 +2198,178 @@ impl<T: Frontend> App<T> {
...
+                                            break tmp_index;
+                                        }
+                                    }
+
+                                    tmp_index = if tmp_index == 0 {
+                                        merged_file_paths_map.len() - 1
+                                    } else {
+                                        tmp_index.saturating_sub(1)
+                                    };
+                                }
```
wongjiahau:
```suggestion
{
                                let marked_file_indices = merged_file_paths_map
                                    .iter()
                                    .enumerate()
                                    .filter_map(|(index, (_, is_marked))| {
                                        is_marked.then_some(index)
                                    })
                                    .collect_vec();
                                *marked_file_indices
                                    .iter()
                                    .rev()
                                    .find(|index| **index < current_index)
                                    .or_else(|| marked_file_indices.last())
                                    .unwrap_or(&current_index)
                            }                                    
```


This part can be simplified by:
1. Precomputing the list of marked file indices (`marked_file_indices`)
2. Using `[find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) and [`rev`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev)



Similar simplification can be done for other match branches (i.e., `Movement::Right =>`, `Movement::Next` etc)


The `marked_file_indices` variable should be declared outside of this `match` expression so that it can also be referred inside `Movement::Right => {...}`.

src/app.rs
```diff
@@ -2197,32 +2198,178 @@ impl<T: Frontend> App<T> {
...
+                                        break current_index;
+                                    }
+                                    tmp_index = tmp_index.saturating_add(1);
+                                    if let Some((_file_path, _marked_file)) =
+                                        merged_file_paths_map.get_index(tmp_index)
+                                    {
+                                        break tmp_index;
+                                    }
+                                }
+                            }
```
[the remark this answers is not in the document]
haxfn: This is a good point and I agree it would be more ergonomic that way. so `<`/`>` navigates exclusively on unmarked open files.

wongjiahau (approved):

PR #1240, opened 2026-01-16
fix(file_explorer): fix Windows

### LINKED ISSUE #1231: can't open file in file explorer menu
Image

select a file name, input `,`, but can't open the file on windows, linux version is ok

Riolku:
Fixes #1231

Due to a combination of generous std::mem::take usage, and hand-written ancestors logic, windows path handling is completely broken, and the error silently clears the internal tree datastructure.


wongjiahau:
Looks good to me!

PR #1264, opened 2026-01-21
feat: Scrolling (eg alt i/k) always centers the screen

SeanWiig:
Right now the half-page scroll action moves the cursor by a half page, which triggers a screen-recenter semi-randomly depending on whether the cursor lands on the last visible line or one past the viewport. This feels choppy. This fix explicitly re-centers the screen on scroll.


wongjiahau:
Thanks @SeanWiig!

PR #1309, opened 2026-01-29
fix(quickfix): first sorted entry different from first received result

wongjiahau:
## Solution
This is fixed by:
1. Order and index the file paths before searching
2. Create a thread that buffers search matches until the first sorted entry is determined

## Evaluation
According to the relevant logs produced after running a few sample global searches, we can see that this implementation is quite performant, where the buffering time takes less than 1% of the total searching time.

> [00:00:02.580] (178207000) INFO   First entry sent after buffering for 2.091458ms (1.5%), ahead of finished time by 131.530375ms
> [00:00:04.098] (16f217000) INFO   First entry sent after buffering for 2.281459ms (1.6%), ahead of finished time by 123.158666ms
> [00:00:35.373] (1710cb000) INFO   First entry sent after buffering for 1.722917ms (1.1%), ahead of finished time by 90.466125ms
> [00:01:02.984] (171b07000) INFO   First entry sent after buffering for 743.375µs (0.0%), ahead of finished time by 61.79425ms

### DIFF
--- a/docs/blog/2026-01-10-evolution-of-delete.md
+++ b/docs/blog/2026-01-10-evolution-of-delete.md
@@ -238,7 +238,7 @@ For example, to delete two selections:
 
 Since most deletions involve only one selection, the menu approach is more efficient for the common case while keeping the Delete Submode available for users who prefer chaining deletions.
 
-## Version 6: Momentary Layer
+## Version 6: Momentary Layer (MoL)
 
 [Momentary Layer](https://zmk.dev/docs/keymaps/behaviors/layers#momentary-layer) combines the best aspects of both the Delete Menu and Delete Submode approaches: it's as efficient as the Delete Menu for single deletions while being as repeatable as the Delete Submode for chained operations.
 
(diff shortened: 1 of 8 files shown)


src/list/grep.rs
```diff
@@ -115,35 +121,50 @@ pub fn run(
...
-            let _ = searcher.search_path(
-                &matcher,
-                path.clone(),
-                sinks::UTF8(|line_number, line| {
-                    if let Ok(locations) = to_locations(
-                        &buffer,
+
+            let matches = {
+                let receiver = {
+                    let (sender, receiver) = std::sync::mpsc::channel();
```
Riolku: Why do we need another sender/receiver pair here?

Riolku: does `search_path` really spawn a new thread? That seems pretty wasteful...?

wongjiahau: Lol I defaulted to thinking callbacks always create a new thread 💀

PR #1330, opened 2026-02-10
refactor(path): change CanonicalizedPath to AbsolutePath

Riolku:
We use CanonicalizedPath generously throughout the codebase. It is quite slow. An absolute path is required for the lexical operations we do, and is overall very convenient especially when we can change directories in the editor. This, combined with the high effort required to refactor all the code to use PathBuf, are good reasons to change this to AbsolutePath.


Riolku:
The only real changes are in absolute_path.rs, the rest are LSP renames

PR #1370, opened 2026-02-25
style: combine marked and save indicators

haxfn:
Previously, the file marked indicator (#) and file dirty (saved/unsaved) status ([*])
were separately showed on the tab bar. Now, we combine both of them:

```
marked  , saved  : [-]
marked  , unsaved: [÷]
unmarked, saved  : [ ]
unmarked, unsaved: [:]
```

Here the horizontal dash in a symbol represents that the file is marked. The colon symbol represents there are unsaved changes in the buffer. Thus something like the divide symbol combines both the horizontal dash and the colon, correspondingly meaning a file with the divide symbol is marked on the tab bar and has unsaved changes.

This looks like this:

image

In addition to the above changes, now, a previous bug is resolved. This bug made us incapable of viewing dirty status of more than one file at a time. This bug was resolved in implementation by bubbling up a Dispatch from buffer, to update Context::file_dirty_status HashMap with key as path name and value as boolean dirty status.


haxfn:
Thanks for the detailed review @wongjiahau given on [#🧩 Feature > ✔ Style: combine marked and save indicators](https://ki-editor.zulipchat.com/#narrow/channel/551672-.F0.9F.A7.A9-Feature/topic/.E2.9C.94.20Style.3A.20combine.20marked.20and.20save.20indicators/with/576205501)

PR #1388, opened 2026-03-02
feat(Space/Picker): Git Branch

haxfn:
Adds the keybind `space k b` which opens a picker listing all git branches.

We use git2 for fetching list of branch names, and we using shell to git checkout 
because shell gives superior error messages that git2: 'xyz file has staged changes, commit them'


src/app.rs
```diff
@@ -2673,6 +2675,32 @@ impl<T: Frontend> App<T> {
...
+        let dropdown_items: Vec<DropdownItem> = repo
+            .branches(None)?
+            .filter_map(|res| res.ok()) // Skip branches that error during iteration
+            .enumerate()
+            .filter_map(|(index, (branch, _type))| {
+                // Transform the branch into a DropdownItem if the name is valid
+                branch.name().ok().flatten().map(|name| {
+                    let branch_name = name.to_string();
+                    DropdownItem::new(branch_name.clone())
+                        .set_rank(Some(Box::from([index].to_vec())))
```
wongjiahau: `.set_rank` is unnecessary here since you're not ranking it based on anything meaningful other than the enumeration index

PR #1405, opened 2026-03-07
fix(insert-mol): should not close non-legend popups

haxfn:


wongjiahau (approved):
Good fix and debugging @haxfn!

PR #1414, opened 2026-03-08
fix(docs): add missing semicolon in flake installation example

willbush:
Noticed a syntax error in the flake install example.


wongjiahau:
Thanks @willbush!

PR #1427, opened 2026-03-09
fix(language/go): return null for unhandled lsp request

j4n-r:
gopls sends this request during startup and stops functioning if it does not receive a response.
Returning null is valid according to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_workDoneProgress_create).
This does not implement the actual intent of the request.


wongjiahau:
Thanks for the fix @j4n-r! Is Go LSP server usable now?

PR #1433, opened 2026-03-11
feat(lang): add latex support

jbedo:
This adds support for latex via the codebook-tree-sitter-latex crate. It's not working yet, I get an error `TopNode::iter: cannot find Treesitter language` when selecting nodes.


jbedo:
Missing syntax highlighting, adding latex to the list of nvim treesitter languages fails with the error:
```
ERROR  [Language::highlight_query]: Falling back to default query; unable to use highlight query of latex from nvim-treesitter due to error: QueryError { row: 167, column: 0, offset: 0, message: "Invalid arguments to set! predicate. Unexpected second capture name @markup.link.url", kind: Predicate }
```

PR #1436, opened 2026-03-13
fix(keymap): process universal keys with translated event

Riolku:
Required for Navigate Forward to register properly in colemak.

A merge error caused the last try of this change to be empty instead.


src/components/editor.rs
```diff
@@ -1701,8 +1701,12 @@ impl Editor {
         context: &Context,
         key_event: KeyEvent,
     ) -> anyhow::Result<Dispatches> {
-        match self.handle_universal_key(key_event)? {
-            HandleEventResult::Ignored(key_event) => {
+        let translated_key_event = context
+            .keyboard_layout_kind()
```
Riolku:
```suggestion
            .keyboard_layout()
```

PR #725, opened 2025-08-08
Adds FSharp LSP & Tree Sitter support

pragun:

PR #1449, opened 2026-03-16
refactor(languages/FSharp): use tree-sitter-fsharp crate

dariof4:
I stumbled upon https://github.com/ki-editor/ki-editor/pull/725#discussion_r2281091248, from what I gathered using a crate instead of pulling from git is more idiomatic so I tried to use the crate and it seems to work now (there have been some recent new releases of https://github.com/ionide/tree-sitter-fsharp, so I guess they must have fixed something while doing them)

P.S.: This editor is really cool, using tree-sitter to make movement through the AST first-class is a really nice idea :-)

### DIFF
--- a/docs/static/app_config_json_schema.json
+++ b/docs/static/app_config_json_schema.json
@@ -106,6 +106,7 @@
         "Odin",
         "CSharp",
         "Clojure",
+        "FSharp",
         "Scala"
       ]
     },


dariof4:
> @dariof4 would you mind to run `just check-config-schema` and commit the updated schema json file?

Done 👍 

> Thank you! I hope you had fun with Ki, feel free to chime in on our Zulip server :)

I saw the Zulip server, I'll definitely hop on there :-)

wongjiahau:
Again, thanks for the contribution @dariof4!

Issue #1458, opened 2026-03-19
bug(SelectionMode/Line): next command skips penultimate line if it's empty


dariof4:
The `next` command skips the penultimate line if it's empty, this doesn't happen in `LINE*` mode

video example (the blinking cursor not showing is only because the asciinema recording doesn't support it, in reality it shows fine):
[![asciicast](https://asciinema.org/a/doNcxBdNS2tLBk3K.svg)](https://asciinema.org/a/doNcxBdNS2tLBk3K)

PR #1464, opened 2026-03-20
fix(selection_mode/LineTrimmed): issues around the last line

### LINKED ISSUE #1457: bug(SelectionMode/Line): can't move back while on the "last line"
Repro: open any file and go to the "last line" (the one after the file ends) while in line mode (need to use next to get to it), at this point you can't go back using the previous command (you can use other commands like left to get unstuck though).

for reference, this doesn't happen in `LINE*` mode.

video example (the blinking cursor not showing is only because the asciinema recording doesn't support it, in reality it shows fine):
[asciicast](https://asciinema.org/a/ZRbWPU3eJZpPrc6d)

dariof4:
Fixes #1457 and #1458, I did a little bit of testing and it seems to work fine, but since I am not experienced with the editor I am unsure if this breaks some more sophisticated behaviour of selecting, I would appriciate some testing.


wongjiahau:
@dariof4 seems like nothing's broken!

PR #1529, opened 2026-04-03
Fix lisp file creation

### LINKED ISSUE #326: Ki doesnt handle urls properly when lsp requests to create a file
the problem is
```rust
Ok(ResourceOperation::Create(create.uri.into()))
```
defined as
```rust
Create(String)
```
[here](https://github.com/ki-editor/ki-editor/blob/master/src/lsp/workspace_edit.rs#L14-L30)
which results in "file:///path/to/file"
this creates it as a subdirectory tree starting at "file:"

this is annoying

`ResourceOperation::Create` should probably take a `PathBuf` or `CanonicalizedPath` instead of a `String` but its a `String` cause of the file explorer and idk how you want to fix this

irevoire:
Fix https://github.com/ki-editor/ki-editor/issues/326

Hey, I used my version of ki with this change for a few minutes, and everything seems to work, but sadly, I can't run the tests, so I don't know if anything else broke.

I'm flooded by the following error: `Error: Path "mock_repos/rust1" does not exist or you don't have access!`

### DIFF
--- a/src/lsp/workspace_edit.rs
+++ b/src/lsp/workspace_edit.rs
@@ -23,7 +23,7 @@ impl TryFrom<lsp_types::ResourceOp> for ResourceOperation {
     fn try_from(value: lsp_types::ResourceOp) -> Result<Self, Self::Error> {
         match value {
             lsp_types::ResourceOp::Create(create) => {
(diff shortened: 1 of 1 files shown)


wongjiahau:
Thanks @irevoire!

PR #1530, opened 2026-04-03
refactor(location_history): make clearing location_history explicit

dariof4:
Fixes #1487 
Fixes #1519

`push_location_history(true)` clears `location_history_forward` which it shouldn't if it gets called by doing `Nav ->`.
To fix this I made it so `push_location_history` never clears `location_history_forward` and added the `clear_location_history` function, which is used in `open_file` (since you do want to clear `location_history_forward` in that case).
I also made it so `location_next` returns at the first match (since `location_history_forward` doesn't get cleared anymore, without this change it would go to the last match, which doesn't make much sense).

I am not 100% convinced this is the cleanest way to go about it though (for example, I made it so `clear_location_history` can be used for clearing both forward and backward location history, but currently `location_history_backward` is never cleared, so I'm not sure if making it work both ways is needed or just makes things more confusing).

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -531,10 +531,13 @@ impl Context {
     pub fn push_location_history(&mut self, location: Location, backward: bool) {
         if backward {
             self.location_history_backward.push(location);
-            self.location_history_forward.clear();
         } else {
             self.location_history_forward.push(location);
         }
+    }
+
+    pub fn clear_forward_history(&mut self) {
+            self.location_history_forward.clear();
     }
 
     pub fn location_previous(&mut self) -> Option<Location> {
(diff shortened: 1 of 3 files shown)


dariof4:
Upon further review it seems that this also fixes #1519 (so I changed the recipe to make it expect to be back on the `main.rs` selection).
I am not sure 100% sure if what I did has the correct behaviour with regards to movement between marked files though.

src/context.rs
haxfn, line 1: Unless there are future use cases for clear_location_history being bidirectional -- in which case you should leave a comment mentioning such a future usecase -- we should just change this function to `clear_forward history` and remove the bool parameter since we currently do not ever clear backward history.

dariof4: I've changed it to `clear_forward_history`, since I can't currently envision a case in which we clear backward history, and well, if there ever is such a case it can always be changed :P
Thanks for the feedback!

wongjiahau:
Thanks @dariof4!

PR #1533, opened 2026-04-06
feat(ux): separate channel for crossterm events

Riolku:
This, combined with select_biased from crossbeam, allows us to prioritize crossterm events when they come in, leading to more responsive UX.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -369,6 +369,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Riolku:
Old: https://asciinema.org/a/e7EeaLHNoqVSSLgm
New: https://asciinema.org/a/BtXbosJtkCUDPFDD

In both cases, I cause a flood of LSP diagnostics.

Edit:

## Steps taken
1. Change the type annotation of the second argument of `Keybinding::new` (`name`) from `String` to `&'static str`.
2. Save
3. Spam Right (`l`) by holding it or repeatedly pressing `l` quickly.

## Expectation
The selection keeps moving forward smoothly without choking.

## Actual (before this PR)
Once Saving, the forward movement will choke due to the UI thread handling a flood of LSP diagnostics notifications.

wongjiahau:
Tested: LSP and crossterm events are still handled as expected.

Thanks @Riolku!

PR #1535, opened 2026-04-07
refactor(ActionGroup): store (action, applied_offset) tuples

### LINKED ISSUE #1470: fix(editor): swap same line sibling nodes not working
Vishal: I think multicursor swap is not working correctly.

`&[("j", Movement::Left), ("l", Movement::Right)]`
`d k k hold-r space release-r j k hold-t l`

expected result:
`&[(Movement::Left, "j"), (Movement::Right, "l")]`
actual:
`&[(Movement::Left, "j"), ("l", Movement::Right)]`

If this is valid, we can add it as a test case somewhere.

[#🛠 Fix > Multicursor swap same line sibling nodes not working](https://ki-editor.zulipchat.com/#narrow/channel/545112-.F0.9F.9B.A0-Fix/topic/Multicursor.20swap.20same.20line.20sibling.20nodes.20not.20working/with/580788549)

dariof4:
Fixes: #1470

Instead of only storing the actions in `ActionGroup`, I made it so it stores tuples of the action and the offset applied to the action, so that while calling `ActionGroup::intersect_with` it applies the negated offsets to retrieve the unnormalized edits to calculate if they intersect correctly.
This should already work, but I'm marking this as draft because there's 3 things I'm not sure about:
1. Is making `ActionGroup` store a single `Vec` with tuples (`Vec<Action, isize>`, like this pr currently does) good or is it better to have `ActionGroup` store `Vec<Action>` and `Vec<isize>` (in this case I think you'd have the issue of keeping them in sync when the actions are sorted).
2. Should there be added a `ActionGroup::new_without_offset` function that takes a `Vec<Action>` and returns an `ActionGroup` with those actions with offset 0, so you dont have to add `,0` to all the actions everytime you call `ActionGroup::new` (or should this be the default behaviour of `ActionGroup::new`, and if needed there could be a `ActionGroup::new_with_offset`)
3. I have to take a look at `ActionGroup::inverse` (does it need to invert the applied_offset?)


wongjiahau (approved):

wongjiahau:
@dariof4 please let me know if you think this is ready to be merged!

wongjiahau:
@dariof4 are you still working on this PR?

PR #1573, opened 2026-04-22
keymap: move Open MoL to `x`

wongjiahau:
This is to solve an RTD issue as described in https://ki-editor.zulipchat.com/#narrow/channel/543759/topic/Rolling.20issue.20with.20Change.20action.20in.20Insert.20MoL/with/584580750

### DIFF
--- a/docs/docs/momentary-layers/open-mol.mdx
+++ b/docs/docs/momentary-layers/open-mol.mdx
@@ -7,7 +7,7 @@ The actions of this Momentary Layer (≡) enters Insert mode.
 
 ## Keymap
 
-<KeymapFallback filename="Insert"/>
+<KeymapFallback filename="Open"/>
 
 ## Tap action: Change
 
--- a/docs/docs/normal-mode/actions.md
+++ b/docs/docs/normal-mode/actions.md
@@ -53,9 +53,9 @@ For example:
 
 <TutorialFallback filename="open"/>
 
-### `Insert`
+### `Open`
 
-Activates the [≡ Insert](../momentary-layers/insert-mol.mdx).
+Activates the [≡ Open](../momentary-layers/open-mol.mdx).
 
 ### `Delete`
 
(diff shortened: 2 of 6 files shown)


dmitry-salin:
As a result of these changes, the bottom left row of keys is filled with MoL symbols. On some keyboard layouts, one of the keys on this row is used to activate the left Alt key (or another modifiers), which is essential for Scroll ↑/Scroll ↓. This means you won't be able to use this row for modifiers on the keyboard firmware level. As a workaround it is possible to map some key for just 'x' for example (to use it as Ki MoL) on some another keyboard layer, but this means that not all editor actions can be made on the default keyboard layer. Example layout: https://configure.zsa.io/voyager/layouts/vyr0a/latest/0

wongjiahau:
@Riolku what we fear has come true, would it be possible to avoid Ki's MoL keys clashing with modifiers placed in custom positions?

wongjiahau:
@dmitry-salin have you tried using homerow mods on your keyboard layout?

dmitry-salin:
> @dmitry-salin have you tried using homerow mods on your keyboard layout?

In my case, it's not a mod, but a built-in feature of the firmware's web configurator. I only use `x` key as an `Alt` key. However, any key can be configured to act as a modifier when held down. The only workaround I can see is duplicating some of the Ki MoL keys on an additional layer, which means that to activate them, I'll also have to hold the key that activates that layer.

wongjiahau:
I see, I'm thinking of making tap x as Open Down and tap X as Open Up, would this compromise work for you?

wongjiahau:
Also, were you not using the Cut MoL which was on `x`?

dmitry-salin:
I'm relatively new to learning Ki and so far there's no conflict with `x`. Using the `X` key as the Open Up key is acceptable, but it requires holding down `Shift`, which is almost the same as the extra layer workaround. I think the extra layer is a more universal solution, as it can resolve conflicts with more than just `x`. Therefore, anyone with a multi-layer keyboard can try changing their layout to make it optimal for Ki. Ideally, the extra layer should contain less frequently used MoL keys.

wongjiahau:
> Therefore, anyone with a multi-layer keyboard can try changing their layout to make it optimal for Ki.

Sounds good!

PR #1585, opened 2026-04-29
fix(editor): highlighting of doc comments

dmitry-salin:
1. Zed uses `comment.doc` in themes and in `highlights.scm`, while nvim uses `comment.documentation`.
2. For doc comments, `tree-sitter-highlight` library generates a sequence of events that results in a kind of nested contexts:
```
HighlightStart (CommentDocumentation)
Source (start: 0, end: 2)
HighlightStart (Operator)
HighlightStart (CommentDocumentation)
Source (start: 2, end: 3)
HighlightEnd
HighlightEnd
Source (start: 3, end: 17)
HighlightEnd
```

This is probably a bug on the `tree-sitter-highlight` side: there's a `HighlightStart (Operator)` event that doesn't correspond to a `Source` event, only a `HighlightEnd`.
In the current code, the third `Source` event and highlight span for comment text are lost, only `///` symbols are highlighted (first and second `Source` events).


wongjiahau:
Thanks @dmitry-salin!

PR #1601, opened 2026-05-08
Add zsh support

irevoire:
Hey!

This PR adds zsh support. Since there is no zsh lsp or formatter, I took the freedom to use the bash one.
That's what Helix has been doing, and it's working ok-ish I think.

Here's an example of it running:

image


wongjiahau:
Thanks @irevoire!

PR #1603, opened 2026-05-09
Make tests crossplatform

irevoire:
And generally more reliable.
Each commit describes precisely each fix.

On this branch, I can run `cargo nextest run` on my macOS.

### DIFF
--- a/src/test_app.rs
+++ b/src/test_app.rs
@@ -19,7 +19,7 @@ use std::{
     cell::RefCell,
     path::PathBuf,
     rc::Rc,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Once},
     time::Duration,
 };
 pub use Dispatch::*;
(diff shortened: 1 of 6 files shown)


src/test_app.rs
```diff
@@ -918,6 +918,12 @@ fn execute_test_helper(
     assert_last_step_is_expect: bool,
     options: RunTestOptions,
 ) -> anyhow::Result<TestOutput> {
+    static SET_ENV_VAR: Once = Once::new();
+    SET_ENV_VAR.call_once(|| {
+        // Safety: Since it's wrapped in a `Once`, it's guarantee that it's never going to be executed twice per process.
+        unsafe { std::env::set_var("RUST_BACKTRACE", "0") };
```
wongjiahau, line 924: Is the use of `unsafe` here necessary?

irevoire: Yes, the function is unsafe and must not be called twice from the same process

PR #1614, opened 2026-05-13
fix: position keymap legend at bottom with dynamic height

descendingnerdliness:
The keymap legend overlay now appears at the bottom of the screen with height that exactly matches its content (buffer lines + title). The tiled layout for other components is adjusted to occupy the remaining space above the keymap band.

Changes in `Layout::recalculate_layout` (`src/layout.rs`):
- Compute keymap height from its editor buffer and title, clamped to terminal height
- Keymap assigned fixed bottom overlay Rectangle; excluded from tiling
- Tiled components use reduced height when keymap is present
- Fixed borrow checker issues and cleaned up unused mut warnings

Keymap remains dynamic (appears on demand, closes with 'esc'). No overflow or extra blank lines.


wongjiahau:
I'm still looking how to improve the code here

descendingnerdliness:
This PR positions the keymap legend as a bottom overlay whose height dynamically matches its content, and makes that sizing robust during terminal resizes.

### `src/layout.rs` — `Layout::recalculate_layout`
- The keymap legend is reserved as a fixed bottom band; the tiled components occupy the remaining space above it.
- The keymap content is refreshed at the full terminal width *before* its height is measured, so the measurement reflects the current width.
- The height is computed from the editor's actual wrapped line count (via `soft_wrap`) rather than the raw buffer line count, accounting for the editor's `width - 1` soft-wrap. This keeps the legend exactly tall enough, so its lines never overflow into the tiled area or leave leftover artifacts above the pane.
- The height is clamped to leave at least one line for the tiled components and never exceeds the terminal height.
- The tiled-rectangle lookup uses a safe fallback instead of panicking if a rectangle is missing.

### `src/components/keymap_legend.rs`
- The keymap content is generated one column narrower than the rectangle. The editor reserves the last column for the cursor and soft-wraps at `width - 1`, so generating at the full width made each keyboard row wrap onto two lines. Generating at `width - 1` keeps each row on a single line.

The keymap remains dynamic (shown on demand, dismissed with `esc`); there is no overflow or blank space.

wongjiahau:
Superseded by #1660

PR #1615, opened 2026-05-14
Add support for gyp files

irevoire:
Hey,

`gyp` files are just json as you can see here: https://github.com/nodejs/node-gyp


wongjiahau:
Thank you!

PR #1621, opened 2026-05-14
fix(dev): config schema generation shouldn't be affected user-defined configs

irevoire:
Hey,

Currently, when updating the configuration ki is mixing up my configuration with the default configuration.
This is because we load the user themes when creating the `RawConfig::default`.
This PR fixes this issue and allows me to run `just check-config-schema` even with a custom theme 😁


wongjiahau:
Thanks!

PR #1624, opened 2026-05-17
feat(keymap): workaround for lack of tap-action on MoLs on non-KPP terminals

dnaq:
Previously when using a non-KPP terminal emulator, MoLs opened up the layer and the layer needed to be manually closed (by pressing escape). However, there did not seem to be any way of actually executing the tap-action of the MoL.

This commit makes it so that double tapping a MoL key executes the tap-action and closes the MoL. E.g. double-tapping `r` will execute the delete action.

This does not affect KPP terminals, since the MoL will be closed as soon as the key is released.

Ran the testsuite, some tests showed up as flaky, but the same tests showed up as flaky without this patch as well.

### DIFF
--- a/src/components/keymap_legend.rs
+++ b/src/components/keymap_legend.rs
@@ -707,4 +707,79 @@ Release hold: Conichihuahua
             ])
         })
     }
+
+    #[test]
+    /// On non-KKP terminals, double-tapping the MoL key (pressing it again while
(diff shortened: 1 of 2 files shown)


dnaq:
Nope, this commit does not work. Manual testing on KPP terminals shows that holding the r-key for too long causes multiple deletes. Sorry for that

PR #1625, opened 2026-05-17
fix(dev): add rustToolchain and rust-analyzer to flake.nix

mario-holubar:
These were missing for me since I don't have the rust toolchain installed system-wide


wongjiahau:
Thanks!

PR #1628, opened 2026-05-18
Add support for glsl

irevoire:
Hey!

This PR adds support for glsl, I used `clang format` for the formatter and from what I see it works pretty well.
I'm not using a C lsp or C tree-sitter because there is some custom one made for glsl specifically.
 
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2331,6 +2331,7 @@ dependencies = [
  "tree-sitter-fish",
  "tree-sitter-fsharp",
  "tree-sitter-gleam",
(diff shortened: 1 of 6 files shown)


package-lock.json
Riolku, line 1: What happened here?

Riolku (changes requested):

package-lock.json
irevoire: Oops, no idea, I guess `just check-config-schema` just did some magic.
I restored the file to its previous state :+1:

Riolku (approved):

PR #1633, opened 2026-05-23
feat(themes): read all supported highlights from zed themes

dmitry-salin:
This adds reading of all supported `Neovim` highlight capture names from `Zed` themes.
Highlights can be divided into categories.

### Exists in `Zed` and `Neovim`, but under slightly different names
|     `Zed`    |        `Neovim`       |
|--------------|-----------------------|
| comment.doc  | comment.documentation |
| string.doc   | string.documentation  |
| string.regex | string.regexp         |

Use this mapping to ensure maximum compatibility with existing `Zed` themes. Read other highlight captures by name exactly as they are defined in `Neovim`.

### Exists in `Zed` and `Neovim` under identical names
- attribute
- attribute.builtin
- boolean
- comment
- constant
- constant.builtin
- constructor
- diff.delta
- diff.minus
- diff.plus
- function
- function.builtin
- function.call
- function.method
- function.method.call
- keyword
- keyword.directive
- keyword.import
- keyword.operator
- label
- markup.heading
- markup.link.url
- module
- number
- operator
- property
- punctuation.bracket
- punctuation.delimiter
- punctuation.special
- string
- string.escape
- string.special
- string.special.path
- string.special.symbol
- tag
- type
- type.builtin
- type.definition
- variable
- variable.builtin
- variable.parameter

### Exists only in `Neovim`
- character
- character.special
- comment.error
- comment.note
- comment.todo
- comment.warning
- constant.macro
- function.macro
- keyword.conditional
- keyword.conditional.ternary
- keyword.coroutine
- keyword.debug
- keyword.directive.define
- keyword.exception
- keyword.function
- keyword.modifier
- keyword.repeat
- keyword.return
- keyword.type
- markup.heading.1
- markup.heading.2
- markup.heading.3
- markup.heading.4
- markup.heading.5
- markup.heading.6
- markup.italic
- markup.link.label
- markup.link
- markup.list.checked
- markup.list.unchecked
- markup.list
- markup.math
- markup.quote
- markup.raw.block
- markup.raw
- markup.strikethrough
- markup.strong
- markup.underline
- module.builtin
- number.float
- string.special.url
- tag.attribute
- tag.builtin
- tag.delimiter
- variable.member
- variable.parameter.builtin

They can be added when porting theme, as schema allows for it.

### Missing in `Zed` and `Neovim`
- syntax.keyword
- syntax.keyword.async
- ui
- ui.bar

It looks like they are not being used and can be removed from `HighlightName` enum.

### Changes to themes
- If theme contains only `comment.doc`, add `string.doc` as well, using the same color. `Neovim`'s language-specific highlight captures use either one or the other
- The colors have not changed


wongjiahau:
@dmitry-salin can you share screenshots of Before vs After? 

I want to know how it actually improves the theming

dmitry-salin:
One example is `Python` documentation strings. `Neovim` queries use `string.doc` for this, and currently there is no way to set its color in theme.

# Before
Before

# After
After

This is applicable to any language that uses `string.doc` in `Neovim` highlight queries.

Overall, I don't see the point of having all `Neovim` highlight capture names in the `HighlightName` enum if there is no corresponding ability to define them within themes.
This just adds full mapping described above. Another possible feature is the ability to override parts of built-in themes via config.

Regarding `markup.*` capture names, I believe it would be preferable to move away from hardcoded mappings to other names in favor of the ability to define them explicitly. These names are mostly used in languages ​​such as `LaTeX` and `Markdown`, I think it would be better to select colors for each theme separately, so that users of that theme can review them.

wongjiahau:
Cool! Is this PR ready to be merged?

dmitry-salin:
Yes, it's ready. The only thing I didn't do was remove the unused variants of the `HighlightName` enum. I didn't find any code paths that set or read it. However, it's better to handle this change in a separate PR to simplify a potential rollback.

### Missing in `Zed` and `Neovim`
- syntax.keyword
- syntax.keyword.async
- ui
- ui.bar

wongjiahau:
Great! I'm looking forward for the follow-up PR :)

wongjiahau:
Thanks!

PR #1644, opened 2026-06-21
fix(lsp/roc): use new formatter command

jwoudenberg:
The Roc language is going through some changes. Among them the invocation of the formatter has changed. This fixes the language configuration to use the new invocation.


wongjiahau:
Thank you!

PR #1646, opened 2026-06-26
fix(grammar): store runtime grammars in xdg cache

TheColorman:
Compiled artifacts like the grammars built by Ki when you do `ki @ grammar build` are currently located in `~/.config/ki/grammars` (`$XDG_CONFIG_HOME`). Since these are not configuration files, but rather non-essential, architecture-specific compiled artifacts, they belong in `$XDG_CACHE_HOME` instead (See the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/#basics)) to prevent cluttering user configuration.

This PR patches `get_runtime_dir` which is currently only used by grammar-related code.


wongjiahau:
Thanks @TheColorman!

PR #1656, opened 2026-07-07
Add release channel metadata

### LINKED ISSUE #1653: Add channel metadata for release discovery and future mise integration
Hi, thanks for building Ki.

I would like to request a small release metadata endpoint that would make Ki easier to install through tool managers such as [mise](https://mise.jdx.dev/).

Since Ki may not be in a stable release phase yet, this does not need to be called `latest` immediately. A channel-based approach would avoid implying stability too early while still giving package managers a stable discovery URL.

For example, GitHub Pages could expose simple channel files:

```text
https://ki-editor.github.io/ki-editor/channels/rc
https://ki-editor.github.io/ki-editor/channels/canary
https://ki-editor.github.io/ki-editor/channels/latest
```

Each file could contain the current tag for that channel:

```text
v0.1.0-rc.2
```

The channel URL may change over time, but the resolved GitHub Release asset should remain immutable, for example:

```text
https://github.com/ki-editor/ki-editor/releases/download/v0.1.0-rc.2/...
```

This could be automated with GitHub Actions:

- when a `v*-rc-*` or `v*-rc.*` tag is published, update `channels/rc`
- when a stable `v*` tag is published, update `channels/latest`
- optionally, when a main/master build is published, update `channels/canary`

This would give mise a stable discovery endpoint. A mise backend can resolve a channel such as `rc` or `latest` to a concrete immutable version, then download the corresponding versioned GitHub Release asset.

Eventually, once a stable channel exists, Ki could be added to the mise registry so users can install it with:

```sh
mise use ki-editor
```

The mise tool name would likely need to be `ki-editor`, because `ki` is already taken, while the installed binary can still remain `ki`.

This would support reproducible installs without requiring the project to declare a stable `latest` release before it is ready.

hh9527:
## Summary

This adds release channel metadata and replaces the old rolling latest build with a clearer stable/snapshot release model.

Closes #1653.

## Design

The release workflow now has three layers:

- A shared release capability that builds the existing platform assets, creates a GitHub Release, updates channel metadata on gh-pages, and performs snapshot GC.
- Stable releases are still automatic: pushing a vX.Y.Z tag creates a normal release and updates stable channel metadata.
- Snapshot releases are manual: maintainers run the release workflow with a branch, tag, or commit SHA, producing an immutable snapshot-YYYYMMDD-HHMM-SHORTSHA prerelease.

Snapshot releases are marked as prereleases so they do not affect GitHub Latest release selection. Each snapshot still has an immutable release tag so package managers can resolve and download a concrete version.

## Release assets

Release assets are now archives instead of raw binaries:

- Linux/macOS: `ki-{os}-{arch}.tar.gz`
- Windows: `ki-windows-{arch}.zip`

Each archive contains the executable under `bin/ki` or `bin/ki.exe`.

This keeps Unix executable permissions inside the tarball, avoids relying on download clients or package managers to add `chmod +x`, and leaves room for future files such as licenses, completions, or man pages.

## Channel metadata

The workflow publishes plain text channel files under gh-pages:/channels:

- stable-latest.txt: current stable version
- stable-versions.txt: append-only stable versions
- snapshot-latest.txt: current snapshot version
- snapshot-versions.txt: retained snapshot versions, newest first

These files are intended as a small channel-discovery layer for package managers and installer tooling. stable-versions.txt can be used by tools that need a stable version list, while snapshot-versions.txt is also the source of truth for snapshot GC.

## Snapshot retention

Snapshot releases are retained by count. The default is to keep the latest 14 snapshot releases.

To change this, update the SNAPSHOT_KEEP variable in .github/workflows/release.yml. For example, SNAPSHOT_KEEP: 30 would retain the latest 30 snapshot releases.

During each snapshot release, snapshot-versions.txt is prepended with the new version and trimmed to SNAPSHOT_KEEP entries. Snapshot releases/tags that are not present in that file are deleted. Stable releases are never garbage-collected.

## Future mise registry shape

Once this lands, a future mise registry entry could use either the GitHub backend or the HTTP backend. The archive layout is intended to make that straightforward because every platform archive exposes the executable under `bin/`.

For example, an HTTP backend-style configuration could look like this:

```toml
[tools."http:ki-editor"]
version_list_url = "https://ki-editor.github.io/ki-editor/channels/stable-versions.txt"
bin_path = "bin"

[tools."http:ki-editor".platforms]
linux-x64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-linux-x64.tar.gz" }
linux-arm64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-linux-arm64.tar.gz" }
macos-x64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-macos-x64.tar.gz" }
macos-arm64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-macos-arm64.tar.gz" }
windows-x64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-windows-x64.zip" }
windows-arm64 = { url = "https://github.com/ki-editor/ki-editor/releases/download/{{version}}/ki-windows-arm64.zip" }
```

The exact registry syntax can be adjusted when submitting to mise, but the important contract is stable: concrete release tags map to immutable archive assets, and each archive contains `bin/ki` or `bin/ki.exe`.

## Testing

Tested on my fork:

- Manually ran the snapshot release workflow.
- Confirmed a snapshot-* GitHub prerelease was created.
- Confirmed release assets were uploaded.
- Confirmed gh-pages contains channels/snapshot-latest.txt and channels/snapshot-versions.txt.

The public GitHub Pages URLs on the fork were not reachable during testing, likely due to fork Pages configuration, but the files were committed to the gh-pages branch.


wongjiahau:
How does the versioning works? Currently we don't have a version system

hh9527:
Good point. This PR does not require Ki to introduce an internal version system yet.

The version here is only the distribution version used by GitHub Releases and package managers:

- Stable releases are opt-in: they only happen when maintainers explicitly push a `vX.Y.Z` tag.
- If Ki is not ready to define stable versions yet, we can simply avoid pushing `vX.Y.Z` tags for now.
- Snapshot releases are independent from stable versioning. They are manually triggered and use an immutable generated version like `snapshot-20260708-1430-abc1234`. (You need to trigger it in https://github.com/ki-editor/ki-editor/actions)

So before Ki has stable versions, we can use only snapshot releases. Once Ki decides to publish the first stable release, pushing a `vX.Y.Z` tag will update `stable-latest.txt` and `stable-versions.txt`.

The gh-pages channel files are mainly a small discovery layer for installers/package managers. Once the first stable release exists, a mise registry entry can point to `stable-versions.txt`, so users could install Ki with something like:

```sh
mise use ki-editor
```

Even without mise, the same channel metadata can support a simple install script. For example, conceptually:

```sh
channel="stable"
platform="linux-x64"

version="$(curl -fsSL "https://ki-editor.github.io/ki-editor/channels/${channel}-latest.txt")"
curl -fsSL "https://github.com/ki-editor/ki-editor/releases/download/${version}/ki-${platform}.tar.gz" \
  | tar -xz
```

This is why the PR keeps the channel metadata separate from Ki internal versioning: the files are just distribution pointers for external tooling, not Ki runtime/application versions.

@wongjiahau

wongjiahau:
> This is why the PR keeps the channel metadata separate from Ki internal versioning: the files are just distribution pointers for external tooling, not Ki runtime/application versions.

Make sense

wongjiahau:
Thanks for creating this PR!

rofrol:
https://ki-editor.org/channels/stable-versions.txt 404

wongjiahau:
@hh9527 would you mind to look into this 404 error? :)

hh9527:
It is expected, and would be resolved after the first snapshot/stable channel is initialized. See #1657. #1658 
@rofrol @wongjiahau

PR #1659, opened 2026-07-09
Add scheduled snapshot releases

### LINKED ISSUE #1657: Publish the first snapshot release with the new release workflow
## Context

PR #1656 added the new release workflow, snapshot releases, archive release assets, and gh-pages channel metadata.

The next step is to publish the first real snapshot release with that mechanism, so users and installer tooling have a concrete stable discovery path to test against.

## Tasks

- Manually run the release workflow to publish the first `snapshot-YYYYMMDD-HHMM-SHORTSHA` release.
- Confirm the release is marked as a prerelease.
- Confirm release assets are uploaded as archives:
  - `ki-linux-x64.tar.gz`
  - `ki-linux-arm64.tar.gz`
  - `ki-macos-x64.tar.gz`
  - `ki-macos-arm64.tar.gz`
  - `ki-windows-x64.zip`
  - `ki-windows-arm64.zip`
- Confirm each archive contains:
  - `bin/ki` for Linux/macOS
  - `bin/ki.exe` for Windows
- Confirm gh-pages channel metadata is published:
  - `channels/snapshot-latest.txt`
  - `channels/snapshot-versions.txt`

## Documentation

Update README.md and the GitHub Pages documentation to show how to install the latest snapshot build.

For example, a simple installer flow could resolve the snapshot channel first:

```sh
channel="snapshot"
platform="linux-x64"

version="$(curl -fsSL "https://ki-editor.github.io/ki-editor/channels/${channel}-latest.txt")"
curl -fsSL "https://github.com/ki-editor/ki-editor/releases/download/${version}/ki-${platform}.tar.gz" \
  | tar -xz
```

## Cleanup

Consider manually deleting older existing releases if they can interfere with GitHub's "Latest" release selection or confuse users.

The old release assets were not produced by the new archive/channel workflow, so they may not be stable installation targets.

hh9527:
## Summary

Adds scheduled snapshot releases so the snapshot channel can initialize and stay updated without requiring maintainers to manually run the release workflow every time.

Closes #1657.

## Design

- Runs the release workflow every day at 19:00 UTC, which is 03:00 Asia/Shanghai.
- Treats scheduled runs as snapshot releases from the repository default branch.
- Keeps the existing manual `workflow_dispatch` snapshot trigger.
- Adds a manual `dedupe` input for snapshot runs:
  - default: `true`
  - `true`: skip publishing if this commit already has a retained snapshot
  - `false`: force publishing a new snapshot for the selected ref
- Keeps stable releases unchanged: stable releases still require pushing a `vX.Y.Z` tag.
- Extends snapshot version suffixes from 7 to 12 git SHA characters before any upstream snapshot release has been published.

## Duplicate prevention

Snapshot dedupe uses the current repository's `gh-pages` branch as the source of truth:

- Reads `channels/snapshot-versions.txt` via the GitHub Contents API
- Checks whether the current commit already appears as the trailing 12-character SHA suffix
- If it already exists, skips build, publish, channel update, and GC

This avoids publishing the same commit repeatedly when there has been no change on the default branch.

If `gh-pages` or `channels/snapshot-versions.txt` does not exist yet, it behaves as an empty snapshot list and publishes the first snapshot.

## Testing

Tested on my fork:

- Merged this workflow into the fork's default branch.
- Manually ran `release.yml` with `ref=master` and `dedupe=true`.
- Confirmed the first run published a snapshot release.
- Confirmed a second manual run with `dedupe=true` became a no-op.

Also ran:

- `git diff --check`


hh9527:
`ci (test)` was cancelled before any steps ran, while the other checks passed. This looks transient. Could a maintainer re-run the failed job? @wongjiahau

wongjiahau:
Thanks!

hh9527:
Great, then first snapshot release will be created after 3:00 tonight, and then snapshot channel will be initialized.

wongjiahau:
We shall wait and see :)

PR #1660, opened 2026-07-10
refactor(layout): position keymap legend via desired_height instead of special-casing

wongjiahau:
## Summary
- Reimplements the approach from #1614 (dynamic bottom-overlay height for the keymap legend) with a less hacky design.
- Adds a `desired_height(width, context) -> Option<usize>` hook to the `Component` trait. Components that want a full-width band reserved at the bottom of the screen (instead of tiling) implement this; the default is `None`, meaning "tile normally".
- `KeymapLegend` implements `desired_height` by reusing its own `display()` method (the same code path used to actually render its content), so there's a single source of truth for its size instead of `layout.rs` reverse-engineering the wrapped-line/title math for a component it doesn't own.
- `Layout::recalculate_layout` now partitions components generically via `desired_height(..).is_some()` rather than hardcoding `ComponentKind::KeymapLegend` checks in three places, and no longer needs to poke `KeymapLegend`'s internal editor rectangle before calling `set_rectangle` to force a fresh-width refresh — `display`/`refresh` take `width` explicitly now, fixing that root cause directly.
- Net effect: `layout.rs` is smaller than in #1614 despite doing the same job, with no component-kind special-casing left in the generic layout algorithm.

## Test plan
- [x] Open ki-editor, trigger the keymap legend (e.g. via the relevant leader key), and confirm it renders as a bottom overlay sized to its content with no overflow or extra blank lines.
- [x] Resize the terminal while the keymap legend is open and confirm the overlay height adjusts correctly (no stale/incorrect height).
- [x] Confirm other windows tile correctly in the remaining space above the keymap legend, and return to full-height tiling once the legend is closed (`esc`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wongjiahau:
@haxfn please feel free to merge this PR if you testing goes well!

haxfn:
Ok.

haxfn:
I've tested it. I think it works fine. My only complaint is:

In the Untitled buffer I can press `space` to open up the space menu
I can `alt+/` to swap focus to Space menu,
But I can't `alt+/` again to swap focus back to the Untitled buffer.

That complaint is unrelated to this PR though. This PR works fine I think. 

Setting to auto merge.

wongjiahau:
Thanks @haxfn!

PR #1679, opened 2026-08-08
Test jj revert

haxfn:
The commands executed:

```
jj new -m 'revert: copy should not drop selection'
# create a new revision

jj bookmark create test-jj-revert
# create a bookmark to be pushed later to remote

jj describe --message 'revert: copy should not drop selection'
# repeating the message, might not be required

jj revert -r c4cc0c3122c31d1b59b07889ef19aa8aa0cbe400 -B @
# reverting (this is the command I got from docs)

jj new ynlnpxtq
# warning message told me resolve conflict in a new child revision

jj resolve
# selecting the one that I want to pick/stay

jj describe --message 'resolve revert conflict'
# just naming the resolving commit/revision.

jj squash
# this was previously suggested by jj warning

jj
jj st

jj edit 9a161512
# switching over to the revision where bookmark is.

jj git push --bookmark test-jj-revert
# pushing to remote
```


haxfn:
I don't understand why the second commit became empty. The resolve commit should have been a separate commit before the revert commit? I thought that was the point of squash: to move the resolving commit before the revert commit.

We don't need to update this branch to remote (origin) master, because the remote master already has revert merged using `git revert` so it would only be confusing to update the branch (i.e rebasing from master).

Closing this pull request.

Further discussion may be done here: [#🌀 General > Learning JJ (Jujutsu) VCS](https://ki-editor.zulipchat.com/#narrow/channel/543761/topic/Learning.20JJ.20.28Jujutsu.29.20VCS/with/615355488)

haxfn:
Another issue is the resolve squash going wrong, or not being reflected:

image

Note there should only be one `events` member initialiser, and the addition should not be present here.

PR #1681, opened 2026-08-24
feat(languages/roc) switch to different tree-sitter grammar

jwoudenberg:
I wrote a tree-sitter grammar for roc specifically intended to provide a good experience writing Roc with Ki. It has a flatter syntax tree which I think suits Ki's tree-based editing commands better. It also includes syntax highlighting.


wongjiahau:
Thank you!

Issue #525, opened 2025-01-19
feature request: auto indent in insert mode


skariel:
when in insert mode, pressing "enter" goes a line down but uses the same column of the line being edited.

for e.g. pressing enter in this state:

![Image](https://github.com/user-attachments/assets/7e931f36-ae6f-4d50-9543-f1de6761319d)

results in this:

![Image](https://github.com/user-attachments/assets/90e993a8-470e-4fa2-88ce-89b83300071f)

however in almost any other editor it would result in the more desirable state: 

![Image](https://github.com/user-attachments/assets/c6f9365e-6acb-48a4-8f37-647aef683348)

PR #1685, opened 2026-08-27
poc(insert): tree-sitter-based auto-indent for Python (#525)

wongjiahau:
## Summary

POC for #525 (auto-indent in insert mode): explores a hand-written `indents.scm` for Python as an alternative to vendoring from nvim-treesitter, which is now archived. Scoped to Python only.

- `shared/src/queries/python/indents.scm`: hand-authored query using Helix's `@indent`/`@outdent` capture convention.
- `Language::indent_query()`: POC accessor, Python only.
- `src/indent_query.rs`: interpreter that runs the query against the buffer's tree-sitter tree.
  - `compute_indent_for_new_line`: used on Enter. Counts one level per distinct line on which an `@indent`-captured ancestor of the cursor begins, and drops one level when the token right after the cursor is `@outdent`-captured (e.g. a closing bracket pushed onto its own line).
  - `compute_reindent_for_outdent_keyword`: used on every keystroke in insert mode (`insert_char`). Fixes the *current* line's indentation in place once it finishes with an `@outdent`-captured token — covers both closing brackets typed by hand, and Python's `elif`/`else`/`except`/`except*`/`finally`.
- `enter_newline()` / `insert_char()`: wire the above into actual edits.

### Why keywords need a different path than brackets

Python's grammar lexes indentation itself (DEDENT/INDENT), so `elif` only parses as part of an `elif_clause` once its line is *already* dedented to align with `if` — which is exactly the correction being computed, a chicken-and-egg problem the tree can't resolve mid-typing. So instead of reading the fix off the (necessarily still-wrong) tree, `keyword_outdent_target` matches the typed text against a keyword list and computes the target as one level less than what a new line after the *previous* line would get (which is unaffected and validly parsed). That keyword list itself is derived straight from `indents.scm`'s own `@outdent`-captured literals (`indent_query::outdent_keywords`) rather than hand-maintained separately, so the query stays the single source of truth.

### Known limitations (see comments in the code)

- A `try:` block missing its first `except`/`finally` is invalid Python, and tree-sitter-python reports `ERROR` nodes for it that can swallow the enclosing `def`/`class`. Helix works around this with `@extend`-based `ERROR`-node patterns, which this POC does not implement — both indentation paths can under-indent while a `try` is in that state.
- A single-line compound statement whose body is already present (e.g. `if x: pass`) will still be treated as opening a scope, since tree-sitter-python's error recovery doesn't reliably distinguish that from "body not written yet" at the node-type level.
- Edits made while already in insert mode force a reparse on every keystroke that goes through `insert_char`/`enter_newline` now (needed for both features to see up-to-date syntax), a change from the original pre-existing "only reparse on leaving insert mode" behavior.

## Test plan

1. Open a `.py` file.
2. Type:
   ```
   def foo():
       if x:
   ```
3. Press Enter at the end of `if x:`.
4. Confirm the cursor lands one level deeper (8 spaces) instead of just copying the `if x:` line's own 4-space indentation.
5. Repeat with `for`/`while`/`class`/`with`/`try`/`except` to confirm the same block-opening behavior.
6. Type `x = 1` on its own line and press Enter — confirm indentation is only copied, not increased (no false positive on plain statements).
7. Type `foo(`, press Enter, then type `)` — confirm the `)` lands back at `foo`'s own indentation, not the hanging-content level.
8. Type:
   ```
   if x:
       pass
   ```
   Press Enter (cursor auto-indents to 4 spaces), then type `elif y:` — confirm `elif` snaps back to column 0 as soon as it's fully typed. Repeat for `else`, and for `except`/`finally` inside a `try` that already has one clause.
9. Open a non-Python file (e.g. `.rs`) and confirm Enter and typing behave exactly as before (no regression).

### DIFF
--- a/shared/src/language.rs
+++ b/shared/src/language.rs
@@ -518,6 +518,17 @@ impl Language {
         })
     }
 
+    /// POC: hand-written `indents.scm` queries, authored in this repo
+    /// (see `shared/src/queries/<id>/indents.scm`), rather than vendored
+    /// from an external source. Currently only implemented for Python
+    /// to validate the approach raised in issue #525.
+    pub fn indent_query(&self) -> Option<String> {
+        match self.tree_sitter_grammar_config.as_ref()?.id.as_str() {
+            "python" => Some(include_str!("queries/python/indents.scm").to_string()),
+            _ => None,
+        }
+    }
+
     fn highlight_query_default(&self) -> Option<String> {
         let config = self.tree_sitter_grammar_config.as_ref()?;
         match &config.kind {
--- a/shared/src/language.rs
+++ b/shared/src/language.rs
@@ -529,28 +529,6 @@ impl Language {
         }
     }
 
-    /// POC: the subset of `indent_query`'s `@outdent`-captured tokens that are keywords
-    /// rather than punctuation (e.g. Python's `elif`/`else`/`except`/`except*`/`finally`).
-    ///
-    /// These can't be told apart from the tree the way closing brackets can: a keyword like
-    /// `elif` only parses as part of its clause once the line is *already* dedented to align
-    /// with the statement it belongs to (Python's grammar lexes indentation itself, via
-    /// DEDENT/INDENT), so the very line whose indentation `compute_reindent_for_outdent_keyword`
-    /// (in `src/indent_query.rs`) needs to correct can't be relied on to already show up in the
-    /// tree as the clause node `indents.scm` captures. This list lets that function recognize
-    /// the keyword from the raw text of the line instead, and compute the target indentation
-    /// from the (unaffected, still validly-parsed) content before it.
-    pub fn outdent_keywords(&self) -> &'static [&'static str] {
-        match self
-            .tree_sitter_grammar_config
-            .as_ref()
-            .map(|config| config.id.as_str())
-        {
-            Some("python") => &["elif", "else", "except", "except*", "finally"],
-            _ => &[],
-        }
-    }
-
     fn highlight_query_default(&self) -> Option<String> {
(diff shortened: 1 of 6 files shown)


fusiongyro:
A good start but not quite perfect, for instance, if you go into insert mode and type:

```
def foo():
    if True:
        print(f"Do something")
```

After you hit enter on that print line, it indents again, which it should not, yielding this:

```
def foo():
    if True:
        print(f"Do something")
            # <- cursor here
```

Should be more like this:

```
def foo():
    if True:
        print(f"Do something")
        # <- cursor here
```

wongjiahau:
Good catch, thanks! Fixed in d3d1b9673.


Verified your exact repro is now fixed:
```
def foo():
    if True:
        print(f"Do something")
        # <- cursor here
```

fusiongyro:
Looks great!

haxfn:
When pressing enter on the line 1 column 0

to create a empty line above.

it unexpectedly indents the whole line.

wongjiahau:
Fixed in 72c69b7.

Root cause: `compute_indent_for_new_line` looks at the tree-sitter node one byte *before* the cursor to find `@indent`-captured ancestors — normally a safe bias to land inside whatever was just typed. But at buffer position 0 (row 0, col 0) there's no byte before the cursor, so it clamped to byte 0 and landed right back inside the `for_statement` starting on the cursor's own line, wrongly counting it as an indent level.

Added a regression test (`enter_newline_auto_indent_python_at_buffer_start`) covering your exact repro:
```
for i in range(1, n):
    print(i)
```
Enter at row 0 col 0 now correctly yields:
```

for i in range(1, n):
    print(i)
```
without the earlier spurious indent.

Issue #719, opened 2025-08-05
Support keyboard layouts making use of dead keys


fnuttens:
I tried implementing [Ergo-L](https://github.com/Nuclear-Squid/ErgoL#English) in Ki but I have no idea what key code to pass for its dead key (`[O]` in QWERTY).

```rust
pub(crate) const ERGOL: KeyboardLayout = [
    ["q", "c", "o", "p", "w", "j", "m", "d", "?", "y"],
    ["a", "s", "e", "n", "f", "l", "r", "t", "i", "u"],
    ["z", "x", "-", "v", "b", ".", "h", "g", ",", "k"],
];
```

wongjiahau:
Is the dead key handled by the keyboard's firmware or by software on the computer?

wongjiahau:
Can you do me a favor by first running Ki, then press the dead key, and check the log file at `ki @ log` for `unhandled event:`?

wongjiahau:
By right, if the event emitted from pressing the dead key reached the terminal app, then you should be able to see `unhandled event:` that correspond to that key in the log file.

fnuttens:
After thinking about it this morning, I doubt the dead key itself was responsible for this log. You see, Ergo-L outputs a `’` when typing `<deadkey><space>`. When I did my test, I first hit `<deadkey>`, then two times `<space>`, followed by `q` to quit the editor.

My theory is as follows:

```
<deadkey>  # nothing happens
<space>    # nothing happens in the TUI, but the `unhandled event: ’` is logged
<space>    # the space menu appears, I can now quit
q          # quit ki
```

To answer your question, I did replace `?` with `’` in the matrix, and it made the editor crash saying this character was unhandled.
I’ll do more tests, first typing `<deadkey>` a single time.

lydell:
> by the ErgoL software

Just in case you didn’t already know, [dead keys](https://en.wikipedia.org/wiki/Dead_key#Electronic_keyboards) are a concept that exists out of the box in many layouts on Windows, macOS and Linux.

For example, the Swedish QWERTY layout has five keys (located on two physical keys) that are dead: https://kbdlayout.info/KBDSW/

[...]

PR #1693, opened 2026-08-30
feat(keymap): support dead-key layouts via base-layout-key, plus docs

wongjiahau:
## What

Adds a "Keyboard Layouts" section to the configuration docs
(`docs/docs/configuration.mdx`), and implements the real, layout-agnostic
fix proposed in the [latest issue comment](https://github.com/ki-editor/ki-editor/issues/719#issuecomment-4106358310).

**Docs:**

- How `custom_keyboard_layouts` works and how to select a custom layout via
  `keyboard_layout`, with an example (using the Ergo-L layout from #719).
- Why Ki historically couldn't bind a command to a dead key's physical slot:
  dead-key composition happens below the terminal (OS/driver layer), so
  terminal apps only ever see the final composed character, never a
  "dead key was pressed" event.
- The existing workaround (previously only documented in an issue
  comment): map the dead key's slot to whichever character it composes
  when pressed alone.
- A link to #719 for readers who want to follow/contribute to the fix.

**Implementation — layout-agnostic keybinding via Kitty protocol's `base-layout-key`:**

- Pins `crossterm` to [vshylov/crossterm#feat/base-layout-key](https://github.com/vshylov/crossterm/tree/feat/base-layout-key)
  (unmerged upstream at [crossterm-rs/crossterm#1074](https://github.com/crossterm-rs/crossterm/pull/1074)),
  which reports the Kitty Keyboard Protocol's `base-layout-key`: the
  physical PC-101 slot pressed, independent of the active OS layout or
  dead-key composition.
- Requests `KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS` alongside the
  flags Ki already enables (`src/frontend/crossterm.rs`).
- Threads `base_layout_code` through `event::KeyEvent`, deliberately
  excluded from its `PartialEq`/`Hash` (mirroring crossterm's own choice)
  so key events built without live terminal data — tests, the `key!`/`keys!`
  macros — keep comparing equal.
- In `KeyboardLayout::make_combined_key_event`, prefers `base_layout_code`
  over the literal composed character for **normal-mode dispatch only**;
  insert mode is untouched, since it never goes through this translation
  and keeps inserting whatever text the terminal actually sends (so
  composed characters like é, ê still type correctly).

**Known limits:**

- Requires an MSRV bump (crossterm main is on 1.85; Ki is currently 1.83) —
  not bumped in this PR since the fork still builds against Ki's current
  MSRV.
- Unix-only — doesn't touch Windows, where KKP is already disabled in Ki.
- Only helps on terminals implementing KKP's alternate-key reporting
  (kitty, foot, wezterm). Alacritty, xterm, tmux passthrough, and similar
  stay unaffected — those users still need the composed-character
  workaround documented above.
- Depends on an unmerged, unreleased upstream crossterm PR — worth
  tracking #1074's progress rather than vendoring the fork long-term.

## Why

#719 asks Ki to support keyboard layouts that use dead keys (e.g. Ergo-L,
Bépo). Investigation in the issue thread found the root cause (dead-key
composition is invisible to the terminal) and a real fix: Kitty Keyboard
Protocol's `base-layout-key`, once crossterm exposes it. This PR both
documents the interim workaround (discoverable where users actually look
for keyboard layout configuration) and lands the actual positional-dispatch
fix behind that upstream field, gated to terminals that support it.

## Test plan

- Read through the rendered Markdown/MDX by eye to confirm formatting,
  code fences, and links are correct.
- Confirmed no other docs page already covered `custom_keyboard_layouts`
  or dead keys (only the JSON schema mentioned the field).
- Added unit tests in `src/components/editor_keymap.rs` covering
  `make_combined_key_event` with and without `base_layout_code` present.
- `cargo build` and `cargo test` pass against the pinned crossterm fork.

---
_Generated by [Claude Code](https://claude.ai/code/session_0135suFU1soxk2QnPJEHG885)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,9 +210,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "base64"
-version = "0.22.1"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+checksum = "ac07cdecf99051d9a5238b80f35af32cdeba5b336e55d957b318b50137e18da5"
 
 [[package]]
 name = "baz-tree-sitter-traversal"
@@ -397,7 +397,7 @@ version = "7.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "958c5d6ecf1f214b4c2bbbbf6ab9523a864bd136dcf71a7e8904799acfe1ad47"
 dependencies = [
- "crossterm",
+ "crossterm 0.29.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-segmentation",
  "unicode-width",
 ]
@@ -505,11 +505,23 @@ version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d8b9f2e4c67f833b660cdb0a3523065869fb35570177239812ed4c905aeff87b"
 dependencies = [
+ "bitflags 2.9.1",
+ "crossterm_winapi",
+ "document-features",
+ "parking_lot",
+ "rustix 1.0.8",
+ "winapi",
+]
+
+[[package]]
+name = "crossterm"
+version = "0.29.0"
+source = "git+https://github.com/vshylov/crossterm.git?rev=0f086ab06bf6d577cdc0972d450d8a2c0dfe9363#0f086ab06bf6d577cdc0972d450d8a2c0dfe9363"
+dependencies = [
  "base64",
  "bitflags 2.9.1",
  "crossterm_winapi",
  "derive_more",
- "document-features",
  "filedescriptor",
  "mio",
  "parking_lot",
@@ -753,7 +765,7 @@ dependencies = [
 name = "event"
 version = "0.1.0"
 dependencies = [
- "crossterm",
+ "crossterm 0.29.0 (git+https://github.com/vshylov/crossterm.git?rev=0f086ab06bf6d577cdc0972d450d8a2c0dfe9363)",
  "pretty_assertions",
 ]
 
@@ -1373,7 +1385,7 @@ dependencies = [
  "convert_case 0.6.0",
  "crc32fast",
  "crossbeam-channel",
- "crossterm",
+ "crossterm 0.29.0 (git+https://github.com/vshylov/crossterm.git?rev=0f086ab06bf6d577cdc0972d450d8a2c0dfe9363)",
  "debounce",
  "diffy",
  "event",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,18 @@ exclude = ['mock_repos/rust1']
 # See:
 # - https://github.com/crossterm-rs/crossterm/pull/956
 # - https://github.com/crossterm-rs/crossterm/issues/828
-crossterm = { version = "0.29.0", features = ["osc52", "use-dev-tty"] }
+#
+# Pinned to a fork carrying the Kitty Keyboard Protocol's `base-layout-key`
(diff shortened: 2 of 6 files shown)


wongjiahau:
Tested this manually and it doesn't seem to fix the dead-key case on macOS.

**Setup**
- macOS 26.4.1, kitty 0.46.2
- crossterm pinned to `vshylov/crossterm@0f086ab06bf6d577cdc0972d450d8a2c0dfe9363` (this PR's fork/rev)
- Built the Ergo-L layout with [Kalamine](https://github.com/OneDeadKey/kalamine) from `keymaps/fr/ergol.toml` in [Nuclear-Squid/ErgoL](https://github.com/Nuclear-Squid/ErgoL), installed the generated `ergol.keylayout` under `~/Library/Keyboard Layouts/`, and selected it as the active macOS input source
- Ki config:
  ```json
  {
    "keyboard_layout": "ergol",
    "custom_keyboard_layouts": {
      "ergol": [
        ["q", "c", "o", "p", "w", "j", "m", "d", "'", "y"],
        ["a", "s", "e", "n", "f", "l", "r", "t", "i", "u"],
        ["z", "x", "-", "v", "b", ".", "h", "g", ",", "k"]
      ]
    }
  }
  ```
  (row 1, slot 9 = `'`, the composed-idle character for Ergo-L's dead key, per the docs workaround this PR adds)

**Repro**
1. Launch `target/release/ki` inside kitty (built from this branch)
2. Enter normal mode
3. Press the physical key at QWERTY's `O` position — on Ergo-L's `base` table (`ergol.toml`) this is the `1dk` trigger:
   ```
   ┃ J   │ M   │ D _ │ ! ¡ │ Y   ┃
   ┃     │   µ │   _ │***¨ │   û ┃
   ```
   i.e. column 9 of the top letter row, marked `***¨` — Ergo-L's master dead key, which composes accents with the following keystroke and outputs `’` (U+2019) when tapped against a bare space (see `[spacebar] 1dk` in `ergol.toml`)

**Observed**
Ki's buffer shows the normal dead-key composition preview (the floating "two dots"/diaeresis indicator), identical to what any non-KKP-aware app would show — not the command bound to `'`. So `KeyboardLayout::make_combined_key_event` never got a `base_layout_code` to prefer; it's falling through to the composed-character path, which is empty/pending mid-composition.

**Hypothesis, not yet confirmed**
On macOS, dead-key composition happens inside Cocoa's native text-input system (`NSTextInputClient`), which owns the keydown *before* an app's own key-handling — including a terminal's KKP encoder — ever sees it. If kitty's macOS backend only receives the final composed text once composition resolves (never a raw keydown mid-composition), it has nothing to derive `base_layout_code`/`REPORT_ALTERNATE_KEYS` from for that keypress, regardless of Ki's own handling. I haven't traced kitty's macOS source to confirm this, and haven't yet cross-checked with `kitty +kitten show_key -m kitty` in isolation (planned as the next step — if that also shows no key event until composition completes/cancels, it'd confirm the gap is in kitty-on-macOS rather than Ki).

If confirmed, this fix would only land on Linux (X11/Wayland kitty/foot, where XKB dead-key composition can apparently be bypassed at the raw-keycode level) and stay a no-op on macOS across all terminals — worth reflecting explicitly in the PR's "Known limits" section so macOS users on Ergo-L/Bépo don't expect this to unblock them. Will report back once `show_key` is checked, and can follow up with kitty upstream if it turns out to be a gap on their end.

syrm:
Hello,

Unfortunately, it doesn't work on Linux (Wayland/Ghostty) either.

wongjiahau:
Yea this is genuinely hard until we have a new terminal emulator standard to read the physical keypress events across different operating systems

syrm:
According to ChatGPT Astra, this isn't possible right now: no terminal supports it, so Ghostty would need to be modified.

syrm:
ChatGPT Astra has created a working version : 
https://github.com/syrm/ghostty/tree/fix/ergol-layout-modifiers
https://github.com/syrm/ki-editor/tree/ergol-layout

wongjiahau:
Can you try creating a PR with what Astra fixed upstream?

syrm:
It won't work without the Ghostty patch… and I don't know if that might cause any problems.

Anyway : https://github.com/ki-editor/ki-editor/pull/1701

PR #1701, opened 2026-09-10
feat: use physical key instead of layout one

syrm:
You need the Ghostty version https://github.com/syrm/ghostty/tree/fix/ergol-layout-modifiers for this to work.
I don't know if it causes any issues.

ChatGPT and Astra made it all work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,9 +210,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "base64"
-version = "0.22.1"
+version = "0.23.1"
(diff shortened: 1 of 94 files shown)


wongjiahau:
Is it working for you?

syrm:
On linux with the patched Ghostty linked yes

wongjiahau:
That's awesome! Have you created a PR on Ghostty based on the patch?

syrm:
No, I don't think AI-generated PRs are allowed in the Ghostty project.

wongjiahau:
> No, I don't think AI-generated PRs are allowed in the Ghostty project.

Hmm... I guess after this is merge we can just mention that it works on this specifically patched Ghostty.


Regarding the patched crossterm, do we really need to include it wholesale here?

syrm:
Yes, it's still necessary

